PR #11, opened 2020-07-29
memory saving and frozen window in B matrix

Liu-Xiaoxiong:
7 files was changed 
__Data.py
__fermisea2.py
__system*  (add  two parameter   ksep and delta_fz)
The problem of tensor D was solved.
Memory saving only implemented in tensor_D tensor_K and AHE. 
Others will cost more time when seperate k points.


stepan-tsirkin:
Finally, resolve the conflicts first.

PR #16, opened 2020-08-14
Optimization of optical conductivity (__kubo.py)

jaemolihm:
Dear developers,

This PR implements the optimization of the optical conductivity routine.

1) The bottleneck was the einsum call, which had many arguments. As Stepan suggested, I refactored this with smaller einsum and matmul calls.
2) The evaluation of Gaussian is optimized by computing only nonzero elements. np.exp(-x) is set to 0 for x > -200.

The runtime without and with the optimizations are the following.
* No optimization (master branch, 3513555): 69.76 sec
* Optimization 1: 8.26 sec
* Optimizations 1 and 2: 5.51 sec

I checked that the symmetric and asymmetric part of the optical conductivity of Fe computed with the old and new codes are identical up to numerical errors (error ~ 1E-12).

I also deleted the "# free memory" parts in __kubo.py. If they should be kept, I will put them again.

Best regards,
Jae-Mo Lihm
Seoul National University

### DIFF
--- a/wannierberri/__kubo.py
+++ b/wannierberri/__kubo.py
@@ -112,7 +112,6 @@ def opt_conductivity(data, omega=0, mu=0, kBT=0, smr_fixed_width=0.1, smr_type='
         else:
             eta = smr_fixed_width # number
 
-        # Hermitian part of the conductivity tensor
         # broadened delta function [iw, n, m]
         if smr_type == 'Lorentzian':
             delta = Lorentzian(delta_arg, eta)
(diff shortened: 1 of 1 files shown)


jaemolihm:
I will remove the WIP tag after I check carefully that results are unaffected for various cases.

jaemolihm:
I fixed a problem of the Gaussian function optimization with adaptive smearing.
Now the code works correctly with Gaussian and Lorentzian broadening, adaptive smearing, and adaptive mesh.

PR #18, opened 2020-08-16
Including a general ws_search_size by changing R=0 coordinate from 62…

manxkim:
… to "origin"

icnt=63 (62 starting from zero) is only for ws_search_size=np.array([1]*3).
If ws_search_size=np.array([a,b,c]), icnt=origin+1 (origin starting from zero), origin=((a*b*c+1)*2+1)/2-1.
Tested for various ws_search_size larger than [1,1,1].


stepan-tsirkin:
Cannot merge now, because the branch was updated but I merged your previous commit using `git cherry-pick`

PR #23, opened 2020-12-20
Examples of SHC

manxkim:
I made two examples calculating SHC of Pt: one is with quantum espresso inputs and the other is with mmn2uHu and other input files. Both of them generate the frequency-scanned and fermi-scanned SHC data.


stepan-tsirkin:
Hi Minsu, I am do not ignore your PR, just I am thinking of a proper concept of storing examples, tests, etc. Storing many files in the repository may make it blow up quickly, especially if those files are updated from time to time.  Maybe we beed to start another repository for examples, which may include all large files.  I will think of this, and suggestions are also welcome. Let this example hang on your branch for some time.

PR #28, opened 2021-02-08
WIP: Optimization of einsum in shc_B_H

jaemolihm:
I optimized `shc_B_H` by replacing `einsum` with matrix multiplication, transposing and reshaping. This improves the performance of opt_SHCqiao calculation.

Optimization of `einsum` reduces the time taken in `shc_B_H` is from 243s to 82s. After the optimization, the bottleneck is `_R_to_k_H` inside `shc_B_H`  (~50% of total time). See details below.

I checked with bulk Pt that the calculation result does not change after optimization.

This optimization was discussed in https://github.com/stepan-tsirkin/wannier-berri/pull/22#discussion_r532286534
I tried 1) using `einsum_path` and 2) factoring out the 'k' index from the einsum call (but still using einsum inside the loop). But, both did not give any speedup.

Dear @migueldiascosta, could you (or the user, Truman Ng Yu - I could not find the GitHub id) try this MR and double check its performance and correctness?

### Summary of the performance test
I used `num_wann=88`, `number of R points = 1163`, `SHCqiao=True`, `len(Efermi)=1`, `len(omega)=1`, `NKdiv=[1 1 1]`, and `NKFFT=[10 10 10]`. I used `kernprof` to profile the code.

If there are ~100 Efermi or omega, `kubo_sum_elements` becomes the bottleneck.
Note that `kubo_sum_elements` scales as O(N_W^2) while `_R_to_k_H` and `_shc_B_H_einsum_opt` scales as O(N_W^3). So, `shc_B_H` becomes more important when `N_W` is larger.

* Before/after optimization, `shc_B_H` takes 90%/76% of time.
* Optimization of `einsum` reduces the time taken in `shc_B_H` is from 243s to 82s.
* For the `einsum` lines, the speedup is from 180s to 20s.
* After optimization, `_R_to_k_H` of `SH_R`, `SR_R` and `SHR_R` takes ~57% of the total time. Optimized einsum takes ~17%. Calculation of `A_H` also takes ~17%.
* Inside `_shc_B_H_einsum_opt`, most (~90%) of the time is spent in the actual matrix multiplication.

### DIFF
--- a/wannierberri/__Data_K.py
+++ b/wannierberri/__Data_K.py
@@ -714,11 +714,25 @@ class Data_K(System):
         return self._R_to_k_H(self.SHA_R.copy(), hermitian=False)
 #PRB QZYZ18, Qiao's way to calculate SHC
 
+    def _shc_B_H_einsum_opt(self, C, A, B):
+        # Optimized version of C += np.einsum('knlc,klma->knmac', A, B). Used in shc_B_H.
+        nw = self.num_wann
+        for ik in range(self.nkptot):
+            # Performing C[ik] += np.einsum('nlc,lma->nmac', A[ik], B[ik])
+            tmp_a = np.swapaxes(A[ik], 1, 2) # nlc -> ncl
+            tmp_a = np.reshape(tmp_a, (nw*3, nw)) # ncl -> (nc)l
+            tmp_b = np.reshape(B[ik], (nw, nw*3)) # lma -> l(ma)
+            tmp_c = tmp_a @ tmp_b # (nc)l, l(ma) -> (nc)(ma)
+            tmp_c = np.reshape(tmp_c, (nw, 3, nw, 3)) # (nc)(ma) -> ncma
+            C[ik] += np.transpose(tmp_c, (0, 2, 3, 1)) # ncma -> nmac
+
     @lazy_property.LazyProperty
     def shc_B_H(self):
         SH_H = self._R_to_k_H(self.SH_R.copy(), hermitian=False)
-        shc_K_H = -1j*self._R_to_k_H(self.SR_R.copy(), hermitian=False) + np.einsum('knlc,klma->knmac', self.S_H, self.D_H)
-        shc_L_H = -1j*self._R_to_k_H(self.SHR_R.copy(), hermitian=False) + np.einsum('knlc,klma->knmac', SH_H, self.D_H)
+        shc_K_H = -1j*self._R_to_k_H(self.SR_R.copy(), hermitian=False)
+        self._shc_B_H_einsum_opt(shc_K_H, self.S_H, self.D_H)
+        shc_L_H = -1j*self._R_to_k_H(self.SHR_R.copy(), hermitian=False)
+        self._shc_B_H_einsum_opt(shc_L_H, SH_H, self.D_H)
         return (self.delE_K[:,np.newaxis,:,:,np.newaxis]*self.S_H[:,:,:,np.newaxis,:] +
             self.E_K[:,np.newaxis,:,np.newaxis,np.newaxis]*shc_K_H[:,:,:,:,:] - shc_L_H)
 #end SHC


stepan-tsirkin:
Thank you Jaemo for amother optimization.  It would be good if someone tests it on more examples before merge. 
We really need to implement automated testing.. 

So, you say that `_R_to_k_H` in `SH_R` , `SR_R` and `SHR_R` takes more than half of the time. But it is also used to get Hamiltonian, get `S_H` and `A_H` (which consists of `A_Hbar` and `D_H`, the latter depends on `V_H`)  Does it mean that all together this routinne takes a vast majority of the time? If so -  that is the new bottleneck. It consists of Foutier transform and matrix rotations, but I guess for large number of Wannier functions the rotations become dominant, because of the numwann^3 scaling that you mentioned.

It would be nice to find a way to optimize this matrix rotation. As an option we may re-write that piece of code in C or Fortran, with calls to Blas. 

Another thing: I dod some attempts to make parallelization inside Data_K class, for case when parallelisation over K-point6s is impossible. Probably  that implementation is slowing down. Maybe cleaning the code from those things cvan help. Previously there was a restriction on the size of minimal FFT-grid and also the memory usage for some quantities was huge. Now we can use arbitrary small FFT grids, and also memory usage was optimized. Therefor probably doing _rotations in parallel is not needed anymore.

stepan-tsirkin:
> I cannot agree more.. Do you have some plans for this?
Not really, I have never done this myself.  do you have any experience?  it is mainly a question  how to choose a proper way of doing it..
> Doesn't np.dot() (and @) call BLAS functions?

Yes, they do, I think. Not sure if we can get much  from using C.fortran. Basically, the overhead of numpy is in crerating and destroying arrays, as well as transpose, but for big matrices this overhead is actually negligible.  Could be that we reached the limit here, but not sure.

>I am curious how this is done.
Also, regarding performance, for a fixed N_k(=N_FFT * N_Kdiv), if one uses a smaller N_FFT, the total time spent in FFT is longer (because FFT scales as O(N_R log N_R)), but with the benefit of having larger NKdiv which is good for parallelization. Is this correct?

It is done easily, as I realized there is no problem. If two R-vectors fall into same place in FFT box - they just add up contributions, and the final result stays the same. Yes, for very small FFT grids, the preformance will go down, so it is a compromise between efficient FFT and efficient parallelisation, and also demand of memory. Usually optimal FFT is same as abinitio grid, but one can go smaller as a compromise, as long as FFT does not become the heaviest part again (like in postw90).

Actually parallelization is another problem. We are still bound to a single node.  We need a better solution than just `multiprocessing` module. THer are some other solutions to study:
https://wiki.python.org/moin/ParallelProcessing

migueldiascosta:
> This optimization was discussed in [#22 (comment)](https://github.com/stepan-tsirkin/wannier-berri/pull/22#discussion_r532286534)
> I tried 1) using `einsum_path` and 2) factoring out the 'k' index from the einsum call (but still using einsum inside the loop). But, both did not give any speedup.

@jaemolihm Once you factor out the 'k' index, the `einsum` calls inside the loop will also use BLAS functions, so that would already be much better than before and similar to what you are doing now (still, you're right, the way you do it now is even faster, by ~10% in the example I'm using)

stepan-tsirkin:
@migueldiascosta ,  so you checked and the results did not change after optimization, right? 
When you confirm that, I will merge

jaemolihm:
> > This optimization was discussed in [#22 (comment)](https://github.com/stepan-tsirkin/wannier-berri/pull/22#discussion_r532286534)
> > I tried 1) using `einsum_path` and 2) factoring out the 'k' index from the einsum call (but still using einsum inside the loop). But, both did not give any speedup.
> 
> @jaemolihm Once you factor out the 'k' index, the `einsum` calls inside the loop will also use BLAS functions, so that would already be much better than before (~10x, in the example I'm using) and similar to what you are doing now (still, you're right, the way you do it now is even faster than that, an additional ~1.1x in the example I'm using)

Hi @migueldiascosta, thank you for the comments.

I am a bit confused. I ran the following script that implements what you suggested (or my understanding of it): https://gist.github.com/jaemolihm/9e0b8ae2852e4b397489586fc068ff99

The function is
```python3
def loop_plus_einsum(C, A, B):
    # Optimized version of C += np.einsum('knlc,klma->knmac', A, B). Used in shc_B_H.
    for ik in range(nk):
        C[ik] = np.einsum('nlc,lma->nmac', A[ik], B[ik])
```

The output is as follows.
```
nk = 100, nw = 50, repetition = 5
einsum        :  6.747 sec
loop + einsum :  6.774 sec
matmul        :  0.240 sec
```

So, the loop + einsum version did not give any speedup. Is the function above what you suggested, or have I misunderstood you?

migueldiascosta:
@jaemolihm try passing `optimize=True` to all the `einsum` calls in that script - it should have no effect on "naive einsum", but it should make "loop + einsum" almost as fast as "matmul"

jaemolihm:
Now I understand. Thanks a lot @migueldiascosta!

stepan-tsirkin:
Thank you @migueldiascosta , I did not know about the `optimize` option. It is useful, although not as good as expanding sums explicitly:

```
In [30]: c.shape                                                                                 
Out[30]: (10000, 20, 20)

In [31]: %timeit np.einsum('klm,kmn,knp->klp',a,b,c)                                             
2.36 s ± 5.15 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [32]: %timeit np.einsum('klm,kmn,knp->klp',a,b,c,optimize=True)                               
210 ms ± 881 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [33]: %timeit np.array([x @ y @ z for x,y,z in zip(a,b,c)])                                   
64 ms ± 182 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

```

stepan-tsirkin:
@migueldiascosta  , do you think it is useful and harmless to put optimize=True to all calls of einsum in the code?
Or is there areason why this option is not set to default in numpy yet?

migueldiascosta:
@stepan-tsirkin for very small matrices the overhead can be high, e.g. see https://github.com/numpy/numpy/issues/11714 and comments

migueldiascosta:
> @migueldiascosta , so you checked and the results did not change after optimization, right?
> When you confirm that, I will merge

@stepan-tsirkin apart from the small numerical differences (~10^-10) that come from using an optimized BLAS library, I get the same results

PR #30, opened 2021-02-10
specify the 3rd comp. of SPN in sXu (mmn2uHu)

manxkim:
Without this, there appears a problem in making sHu and sIu when NB_out /= NB_in.


stepan-tsirkin:
thanks for the fix

PR #31, opened 2021-02-15
shc_alpha, shc_beta, shc_gamma added

manxkim:
The new parameters, also used in Wannier90, are added to the set of additional parameters.
There is also going to be a PR in -org.


jaemolihm:
It seems that one *cannot* use the previous scheme of computing all 27 elements at once. Is this intended?

manxkim:
Thank you for letting me know this. Not intended and I am going to add it.

stepan-tsirkin:
Hi Minsu, it looks like the fermiOcean features disappear in the __integrate.py file with this PR.  Probably you missed it while resolving conflicts.could you correct it.

manxkim:
During merging I unintendedly deleted that. I return this.

PR #34, opened 2021-03-09
Shiftcurrent

Julen-ia:
An option for the calculation of the shift current has been implemented. It has been tested in GaAs, and the results agree with those obtained by Wannier90.


wannierberri/readme.tmp
```diff
@@ -0,0 +1 @@
+kubo is called from integrapte.py, so shift current should also be called from there
```
stepan-tsirkin, line 1: this file is not needed.

stepan-tsirkin (commented):
Hi Julen, 
thanks for implementation.
I see only that your implementation should not break anything already existing, and I believe that you benchmarked the results with postw90.x . Did you test the performance? Is it faster than postw90? Or does it depend on parameters, like the number of frequencies/Fermi levels/Wannier functions/points in ab initio grid? You are using np.einsum in many places, did you make sure it does not create problems for speed? You may check it by running 

`python3 -m cProfile example.py`

Regards,
Stepan.

wannierberri/readme.tmp
Julen-ia: Hi Stepan

I performed tests on GaAs, it looks much faster than postw90. I ran WB with one processor in my local machine, fine mesh of NKdiv=[7 7 7], NKFFT=[13 13 13], NKtot=[91 91 91] one Fermi energy, 200 frequencies, and it took 972 s. 

I ran postw90 in same 91x91x91 k-grid and same frequencies, in parallel with 8 processors, it took 1772 s. If Im not wrong, this means WB is more than one order of magnitude faster per processor. 

But it is possible that some of the np.einsum could be speed up by some other means.

PR #37, opened 2021-03-22
Automatic tests, continuous integration

jaemolihm:
I made automatic test suite using pytest. Currently, there are two tests.
1. Test of `Data_K`. Calculate several fields using slow FT and FFTW and compare the results.
2. Test of Kubo integration. Compare *.dat output with reference output in `tests/reference`.

After this gets merged, I will write somewhere (where?) how to generate tests for other features.

To avoid having large files, I compressed the `mmn` and `amn` files. Also, the `sHu` and `sIu` files are not stored and computed when running test using `mmn2uHu`.

Also, I enabled continuous integration test using GitHub actions. When a commit is pushed to the GitHub repository, automatic test is run.
By clicking the "Actions" menu, one can see the output. For example, in my fork:
https://github.com/jaemolihm/wannier-berri/actions

- [ ] The reference  data for SHC should be updated if Minsu's bugfix is merged.


stepan-tsirkin:
Thank you for implementing this. As far as I already merged Minsu's bugfix, could you update the reference before merging this PR. 

You may write instructions to implement more tests in a separate file in the root directory, named like README_TESTS.rst or similar.

jaemolihm:
I updated the reference data. This PR can be merged now.
I also enabled calculating code coverage and uploading to codecov, which is useful to find out which part of the code is not tested.

I will write README_TESTS.rst soon.

jaemolihm:
If you want, you can go to https://codecov.io/gh/stepan-tsirkin/wannier-berri/settings/badge and copy and paste it into `README.rst` to have the code coverage badge.
The link can be accessed only by the owner of the repository.

PR #40, opened 2021-04-11
Reverse index of tenser D and K

Liu-Xiaoxiong:


Liu-Xiaoxiong:
Change the way of changing 'derivative' and 'berry curvature/morb' index of D/K

Liu-Xiaoxiong:
Done! and change sth in tab_plot.py

PR #42, opened 2021-04-15
switch to numpy if there is a problem importing pyfftw

stepan-tsirkin:


wannierberri/__utility.py
```diff
@@ -17,12 +17,18 @@
...
 import numpy as np
-import pyfftw
+
 from lazy_property import LazyProperty as Lazy
 from time import time
 from termcolor import cprint 
 import functools,fortio,scipy.io
 
+try: 
+    import fftw
```
jaemolihm, line 27: Shouldn't this line be `import pyfftw`?

stepan-tsirkin: oups, thank you! 
This is a sign that I should not anymore merge my own PRs without reviews by others. You may include it into your next PR.

PR #44, opened 2021-04-18
Fix formatted read of Wannier90 files

### LINKED ISSUE #39: Inconsistency in reading UXU
There seems to be an inconsistency in reading the unformatted and formatted UXU files.

I generated an unformatted and a formatted uHu file from the same wavefunctions. And then I read them using `class UXU`. Below is the result. The two UXU data becomes identical only after swapping two nnb indices.

```
>>> import numpy as np
>>> from wannierberri.__w90_files import UXU
>>> a = UXU("Fe")
----------
  uHu   
---------
formatted == False
using fortio to read
reading Fe.uHu : <Created on  2Apr2021 at 18:17:33>
----------
 uHu OK  
---------

>>> b = UXU("Fe_formatted", formatted=True)
----------
  uHu   
---------
formatted == True
reading Fe_formatted.uHu : <Created on  2Apr2021 at 18:17:11>
----------
 uHu OK  
---------

>>> np.linalg.norm(a.data - b.data)
6981.984517351669
>>> np.linalg.norm(np.swapaxes(a.data, 1, 2) - b.data)
7.742781712282713e-08
```

Maybe the last line of the following code should be `self.data[ik,ib2,ib1]`, not `self.data[ik,ib1,ib2]`?
https://github.com/stepan-tsirkin/wannier-berri/blob/4bd29e4a40b4d2fa6da5e2c7ba11947c2d9c8649/wannierberri/__w90_files.py#L381-L386

jaemolihm:
As discussed in #39, I fixed the reading of formatted UHU and SHU files.
Other changes:
* Fixed typo in the reading of formatted SPN files (`xrange` -> `range`).
* Fixed the writing of formatted SXU files in mmn2uHu. (The spin index should not be the fastest index.)
* Added comments in `__w90_files.py` explaining the content of data arrays.
* Added test for reading formatted files
* Added GaAs Wannier90 data
* Fixed the typo `fftw` -> `pyfftw` mentioned in https://github.com/wannier-berri/wannier-berri/pull/42#discussion_r615267849.

>> Maybe to the code more readable, we could read all arrays assuming F-order, and after reading everything take the transpose.

> Probably you are right here, you may do this in a PR.

I decided not to do this because I realized doing this can make the code more complicated. Instead, I added more comments explaining the content of temporary arrays.

I reopened to make pseudopotential files be in a compressed format.

Closes #39.


stepan-tsirkin (approved):
looks fine

stepan-tsirkin:
Hi @jaemolihm ,
only one question - do you understand why the codecov checks are never finished? Is there smth wrong with those checks?

jaemolihm:
Hi @stepan-tsirkin , I guess the reason is that the codecov was not setup in the previous PR, so it does not have information about current master branch. Now, the codecov works for every push and PR, so in later ones, it will cork.

So let's wait and see what happens in the next PR.

PR #48, opened 2021-05-05
 fixed a bug with low-dimensional TB models

stepan-tsirkin:
Fix of a problem reported by Se Young Park on the mailing list.

### DIFF
--- a/tutorial/Li_bcc_PythTB2WB.py
+++ b/tutorial/Li_bcc_PythTB2WB.py
@@ -33,7 +33,7 @@ for R in ([1,0,0],[0,1,0],[0,0,1],[1,1,1]):
     my_model.set_hop(t, 0, 0, R)
 
 
-system=wb.System_PythTB(my_model,getAA=True)
+system=wb.System_PythTB(my_model,berry=True)
 Efermi=np.linspace(-7,16,5000)
 # Define the generators of the point group of the crystal (Im-3m)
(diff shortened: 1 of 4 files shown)


jaemolihm (commented):
Looks good, except one question. If `periodic` is not provided at all, it becomes `[True, True, True` (due to the `default_parameters`). Doesn't this behavior contradict "If less then 3 values provided, the rest are treated as ``False``"?

An option would be to force the argument `periodic` to be always of length 3.

stepan-tsirkin:
I think there is no contradiction. If the argument `periodic` is not provided at all, it is taken as default `[True,True,True]`, i.e. as 3 values provided. If one provides `periodic=[]`, it is taken as `[False,False,False]`. Although, not sure if using wannierberri for non-periodic systems ever makes sense.

stepan-tsirkin:
I mean "zero values provided" and "no values provided" are different things. Do you agree? Or is it confusing and needs more explanatory docstring?

jaemolihm:
I see. Maybe adding "If not provided, set to ``(True, True, True)``." or something similar would remove the confusion.

stepan-tsirkin:
Ok, I added the printing of default parameters to the docstrings. Now should be clear. 
I also set the frozen_max to -Inf , because  is nothing is known about frozen window, it should not be used for the orbital moment

jaemolihm (approved):

PR #49, opened 2021-05-11
Print K point progress

jaemolihm:
This MR implements printing of the progress for K point processing.

Since `map` does not allow to monitor the number of remaining jobs, I instead used `imap`.

The progress is printed every `nstep_print` K points calculated. The step size is decided as `nstep_print = max(1, nproc, numK // 100)`. If `numK` is small, the step size is `nproc`. If `numK` is large, ~100 steps are printed.

### `nprocs = 0`, `numK = 9`
```
processing 9 K points :
# K-points calculated  Wall time (sec)  Est. remaining (sec)
                   1              3.5                  27.9
                   2              6.8                  23.8
                   3             10.4                  20.8
                   4             14.2                  17.8
                   5             17.8                  14.3
                   6             21.2                  10.6
                   7             24.7                   7.1
                   8             28.3                   3.5
                   9             31.7                   0.0
time for processing      9 K-points  on   0 processes:    31.7072 ; per K-point          3.5230 ; proc-sec per K-point :          3.5230
```

### `nprocs = 2`, `numK = 9`
```
processing 9 K points :
using a pool of 2 processes
# K-points calculated  Wall time (sec)  Est. remaining (sec)
                   2              3.8                  13.3
                   4              7.5                   9.4
                   6             11.0                   5.5
                   8             14.6                   1.8
time for processing      9 K-points  on   2 processes:    17.8403 ; per K-point          1.9823 ; proc-sec per K-point :          3.9645
```

### `nprocs = 2`, `numK = 625`
```
processing 625 K points :
using a pool of 2 processes
# K-points calculated  Wall time (sec)  Est. remaining (sec)
                   6              0.1                  12.4
                  12              0.2                  11.4
                  18              0.3                  11.2
                  24              0.4                  11.0
                  30              0.5                  10.7
                  36              0.6                  10.5
...(omitted)
                 618             10.9                   0.1
                 624             11.0                   0.0
time for processing    625 K-points  on   2 processes:    11.2981 ; per K-point          0.0181 ; proc-sec per K-point :          0.0362
```


wannierberri/__evaluate.py
```diff
@@ -27,29 +27,51 @@
...
-        nproc_=nproc
+        nproc_ = nproc
+
     if not (symgroup is None):
         res=[symgroup.symmetrize(r) for r in res]
     for i,ik in enumerate(selK):
         K_list[ik].set_res(res[i])
     t=time()-t0
-    print ("time for processing {0:6d} K-points  on {4:3d} processes: {1:10.4f} ; per K-point {2:15.4f} ; proc-sec per K-point : {3:15.4f}".format(len(selK),t,t/len(selK),t*nproc_/len(selK),nproc) )
+    print ("time for processing {0:6d} K-points  on {1:3d} processes: {2:10.4f} ; per K-point {3:15.4f} ; proc-sec per K-point : {4:15.4f}".format(
```
stepan-tsirkin, line 73: It may be confusing  to write `"on 0 processes"` ?
maybe modify this line to write `"in serial"` when `nproc==0`

stepan-tsirkin (commented):
Looks good.  
just one question - the [docs](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.imap) say that  `imap` is `A lazier version of map().` Do you understand what it means? Did you check if evaluateion time did not change? I guess so, because in our case evaluation of the K-point is much heavier than the overhead of parallelization, but just in case.

wannierberri/__evaluate.py
jaemolihm: Done.

jaemolihm:
I guess "lazier" means that in `imap`, the jobs (K points) are sent to processes one by one, while in `map`, the jobs are first send in large chunks, and then in smaller chunks.

https://bugs.python.org/issue40110
> "Lazy" has several possible aspects, of which Pool.imap() satisfies some:
> - Its iterable argument is materialized one object at a time.
> - It delivers results one at a time.

Regarding performance, there is a small but observable performance penalty due to the parallelization overhead. I tested with very a small system size, `numK=10000` and `nproc=30`. The reported times are the proc-sec per K-point in ms. Results are average and standard deviation of 5 runs.

For the cluster I used, the overhead was important when less than 20ms was spent in each chunk.
- `map`: 5.3 ± 0.2
- `imap`: 18.9 ± 2.7
- `imap`, `chunksize=2`: 8.7 ± 0.9
- `imap`, `chunksize=4`: 5.5 ± 0.2

For larger jobs, there is no performance penalty.
- `map`: 261.4 ± 18.1 
- `imap`: 249.6 ± 15.2

An option to avoid the performance drop with the very small job case is to set the `chunksize` parameter. For example, by setting `chunksize = max(min(int(numK / nproc / 100), 10), 1)`, the performance penalty due to `chunksize` is less than 1% (because at least there are 100 chunks in all processes).

What do you think?

stepan-tsirkin:
I guess, the worst scenario for small `chunksize` is a tight-binding toy model with only nearest-neighbour hoppings. Then the number of `R` vectors is small, and hence the FFT grid is small, and the number of K-points may be enormous ~10^6 .  with `chunksize=10`, the overhead becomes ~half hour. Well, maybe the calculation will go to many hours in that case anyway.

On the other hand, as I understand the biggest problem of big `chunksize` is the memory, is that right? So, for big systems we need `chunksize=1`

I think `chunksize ` should be a free parameter of the `integrate()` and `tabulate()` functions.
And the default may be set to the formula that you suggest.  We may also print a message for those who do not read docs , smth like 

`"chunksize={} used for parallelization".format(chunksize)+ ("Smaller values recommended for large systems" if chunksize>1 else "") +  "Larger values may be used for small systems.  Read the docs for more info" `.

or similar. 

Note, the multiprocessing is limited to one node, and it is becoming a severe limitation.  therefore I am thinking how to substitute it, see  #50 . But that is a separate question.

jaemolihm:
I added `chunksize` as the parameter of `integrate` and `tabulate`, and added the output
```
chunksize = 1 used for parallelization. Larger values may be used for small systems. Read the docs for more info.
```
or
```
chunksize = 3 used for parallelization. Smaller values recommended for large systems. Read the docs for more info.
```

Memory is not the limiting factor. Since all the input and output (K point and results, respectively) are saved as a list anyway, chunksize does not increases memory usage. In a given process, each K point is treated one by one, even when `chunksize > 1`. So there will be e.g. one Data_K object in the memory of one process (up to garbage collection).

The problem with `chunksize > 1` is that it increases granularity for parallelization. For example, if there are only 10 K points and `chunksize = 10`, only 1 process will do all the work, irrespective of `num_procs`.

I agree that the transition to other parallelization scheme is desirable.

stepan-tsirkin (approved):
Ok, looks fine now

Issue #50, opened 2021-05-11
Improving parallelization


stepan-tsirkin:
The big limitation of the present implementation is that evaluation is possible only on one node.   That is a limitation of the `multiprocessung` module. 

What are the other options in terms of ease of implementation and/or efficiency?
Does anyone have exprience on that?

After some search, I have an impression that  [Ray](https://ray.io/) is a good choice. 

Any thoughts/suggestions?

jaemolihm:
As the ray branch is merged, we may close this issue.

PR #52, opened 2021-05-13
Parallelisation with Ray

stepan-tsirkin:
Hi, following #50 , I created the pull request just to open a discussion of implementation. It kind of works, but not sure if it is optimal in terms of memory. In particular I get the following message 

```
2021-05-13 22:34:18,586	WARNING worker.py:1115 -- Warning: The remote function wannierberri.__evaluate.paralfunc has size 12812372 when pickled. It will be stored in Redis, which could cause memory issues. This may mean that its definition uses a large array or other object.
```

Maybe it is still not worse than what we have with multiprocessing, where everything is also pickled. But probably may be improved.

I did not test on a multi-node cluster yet, but `tutorial/parallel-ray` contains a toymmodel code that works on slurm

The print of estimated time does not work on ray, at least with the present implementation.

Any suggestions?


stepan-tsirkin:
I added the requirement for python 3.6 her , following #54  @jaemolihm

jaemolihm:
When I try to install, I get an error. Is `ray>=1.3` in setup.py correct?
```
  Could not find a version that satisfies the requirement ray>=1.3 (from wannierberri==0.9.2.3) (from versions: 0.1.1, 0.1.2, 0.2.0, 0.2.1, 0.2.2, 0.3.0, 0.3.1, 0.4.0, 0.5.0, 0.5.2, 0.5.3, 0.6.0, 0.6.1, 0.6.2, 0.6.3, 0.6.4, 0.6.5, 0.6.6, 0.7.0, 0.7.1, 0.7.2, 0.7.3, 0.7.4, 0.7.5, 0.7.6, 0.7.7, 0.8.0, 0.8.1, 0.8.2, 0.8.3, 0.8.4, 0.8.5, 0.8.6, 0.8.7, 1.0.0rc0, 1.0.0rc1, 1.0.0rc2, 1.0.0, 1.0.1, 1.0.1.post1)
```

stepan-tsirkin:
@jaemolihm , the tests say "cancelled after 360m" it is obviously a problem, which should be solved.  But can we put a lower threshold, that if tests do not end in , say ~30 min, they should be stopped ?

jaemolihm:
> @jaemolihm , the tests say "cancelled after 360m" it is obviously a problem, which should be solved. But can we put a lower threshold, that if tests do not end in , say ~30 min, they should be stopped ?

You can add a line `timeout-minutes: 30` to `.github/workflows/ci.yml`: see https://github.com/jaemolihm/wannier-berri/blob/timeout/.github/workflows/ci.yml#L11

stepan-tsirkin:
@jaemolihm 
>When I try to install, I get an error. Is ray>=1.3 in setup.py correct?

Well, [1.3.0](https://pypi.org/project/ray/) it is the latest version. Not sure if it is necessary, but it should be available

stepan-tsirkin:
Ok, it works, no time to test performance on a multi-node cluster

stepan-tsirkin:
I checked, it works, on two nodes, and makes a speedup of a factor almost 2 compared to one node. check the new tutorial. It is not counting an overhead of initializing a ray cluster on slurm (there are long sleeps needed), but that ammounts to a minute maximum, so negligible for really heavy jobs. Still not sure how it is with memory efficiency, but I think we can merge at this point and upload to pip, and wait for feedback from users.

@jaemolihm , @Liu-Xiaoxiong , what do you think?

jaemolihm:
I will try this branch and give some feedback.

One question: it seems for ray, the `nproc` argument is ignored and `nparK=int(round(ray.available_resources()['CPU']))` is used. Is this correct?
1) I think `nproc` argument should not be ignored. Often I run wannier-berri in a desktop, and I do not want it to consume all CPUs...
2) The automatically chosen `nparK` seems to be the number of logical cores, not the number of physical cores. It would be better to use the latter.

One could use `psutil.cpu_count(logical=False)` as the default. This also applies to `multiprocessing`.

```
processing 349 K points : using a pool of 8 processes.
```
```python3
jmlihm@DESKTOP-K0F0FN5:~/program/wannier-berri/wannierberri$ python3
Python 3.6.7 (default, Oct 22 2018, 11:32:17)
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import psutil
>>> psutil.cpu_count(logical=False)
4
>>> psutil.cpu_count(logical=True)
8
```

jaemolihm:
`parallel_module='ray', numproc=0` initializes ray, but the calculation is run in serial.

tests/test_integrate.py
```diff
@@ -151,3 +158,9 @@ def test_Fe_sym_refine(check_integrate,system_Fe_W90_sym, compare_energyresult,q
                   suffix="sym" , suffix_ref="sym", Efermi=Efermi_Fe , comparer=compare_energyresult )
 
 
+def test_Fe_parallel_multiprocessing(check_integrate, system_Fe_W90, compare_energyresult,quantities_Fe,Efermi_Fe):
+    """Test anomalous Hall conductivity , ohmic conductivity, dos, cumdos in parallel with multiprocessing"""
+
+def test_Fe_parallel_ray(check_integrate, system_Fe_W90, compare_energyresult,quantities_Fe,Efermi_Fe):
```
jaemolihm, line 164: `test_Fe_parallel_multiprocessing` is empty.

stepan-tsirkin: yes, that was bad resolving of conflicts. fixed

stepan-tsirkin:
@jaemolihm 
>One question: it seems for ray, the nproc argument is ignored and nparK=int(round(ray.available_resources()['CPU'])) is used. Is this correct?

Actually, with `ray` neither is used, `ray` manages resources by itself. the nparK is just to have something to print. Also, at the moment there is confusion between nparK and numproc, because they are meant the same. I was thinking of additional parallelization inside the K-point evaluation, but I am not sure if it actually works at the moment.

yes by default it occupies all available resources, and it may be advantagous, e.g. to limit number of CPUs per node to use more memory per process. Not sure if `ray` may handle that automatically , e.g. if memory is not enough, the evaluation of a remote would wait until other processes finish, or may it lead to hang up. 

I need to read the docs about it. Other (or complementary) option is to limit resources at the level of the sbatch script. 

> parallel_module='ray', numproc=0 initializes ray, but the calculation is run in serial.

I do not think so, because at the moment with `ray` the `numproc` (`nparK`) is reset to the number of available CPUs., Yes, we need to think of more meaningful set of default parameters for `nparK` and `parallel_module`

tests/test_slurm.py
```diff
@@ -0,0 +1,41 @@
...
+     '--partition',
+     'express',
+     '--command',
+     "'python -u wb-example.py'",
+     '--num-gpus=3',
+     '--sleep-head',
+     '123.45',
+     '--sleep-worker=67',
+     '--no-submit']
+    , capture_output=True )
```
jaemolihm, line 21: `capture_output` was introduced in python 3.7. https://docs.python.org/3/library/subprocess.html. To use this feature, we need to update `python_requires` in `setup.py`.

It would be good to add tests for all supported python versions. To do so, we need to modify `.github/workflows/ci.yml`, line `python-version: [3.8]` to e.g. `python-version: [3.7, 3.8]`.

stepan-tsirkin: Done, checks are passed both for 3.7 and 3.8

stepan-tsirkin: done! tests are passing for 3.7 and 3.8

stepan-tsirkin:
One weird thing I was observing. If I include 'init ray' in the top of the `__evaluate.py` , then the non-parallel tests would never end. (both locally and on github) However, making a normal run of the code was working ok.That's why I made `ray init` only once, and only when it is really needed, and then I am passing the reference to the model as parameter to a function. That looks like a conflict of pytest and ray, or my misunderstanding of how they actually work.

stepan-tsirkin:
After studying `ray` i find it much better than multiprocessing, in particular because it allows to put large objects (system) to the object store only once, thus saving memory. We can also pass it only by reference, so the matrices like `AA_R` in `Data_K` may be turned into `lazy_propertiy`, i.e. they will be evaluated only when/if needed.  

Also Ray uses a higher level of serialization, so no need for this ugly construction with defining a top-level function

https://github.com/wannier-berri/wannier-berri/blob/000479604917ef2a1874912d90366044adb201f7/wannierberri/__evaluate.py#L208-L210

and then making a partial function of it

https://github.com/wannier-berri/wannier-berri/blob/000479604917ef2a1874912d90366044adb201f7/wannierberri/__evaluate.py#L123-L124

So, I wonder if it is necessary to maintain the `multiprocessing` parallelisation in eval_K. removing it, will make it easier the further development. 

Or is there a risk that `ray` is not well-established, and may fail to work on some systems?

jaemolihm:
The only problem I can think of in removing `multiprocessing` support is usage in non-slurm clusters (e.g. PBS torque). But since `multiprocessing` supports only single node, if ray can be used within a single node without any tweak of the submission script, then this is not a problem.

stepan-tsirkin:
On a single node, ray runs with out  special submission scripts. So, if we change to ray, a user may even not notice the difference. 

All this long script is needed to start ray on several nodes, and connect it into a cluster. Because, in principle, ray allows to connect any computers into a cluster, e.g. my desktop with your laptop.

 I think, similar things can be done with PBS. here is an [example ](https://github.com/ray-project/ray/issues/12216), not sure if it works. @jaemolihm , do you have a PBS cluster at hand to try it?

jaemolihm:
The following script worked for a cluster using PBS Torque. It's not as complicated as I expected 😄 

I used `parallel_module='ray-slurm'`. The option could be renamed to `ray-cluster` or `ray-multinode` or something similar.

```bash
#!/bin/bash
#PBS -N test
#PBS -q batch
#PBS -l nodes=2:G2:ppn=2
#PBS -l walltime=00:10:00
#PBS -e error
#PBS -o output
cd $PBS_O_WORKDIR

# This script is taken and modified from https://github.com/ray-project/ray/issues/10466#issuecomment-737476909

jobnodes=`uniq -c ${PBS_NODEFILE} | awk -F. '{print $1 }' | awk '{print $2}' | paste -s -d " "`

redis_password=$(uuidgen)
export redis_password

thishost=`uname -n | awk -F. '{print $1.}'`
ip=`hostname -i`
port=6379
ip_head="${ip}:${port}"
export ip_head
echo "IP Head: $ip_head"

echo "Allocate Nodes = <$jobnodes>"

echo "set up ray cluster..."
for node in `echo ${jobnodes}`; do
    if [[ ${node} == "${thishost}" ]]; then
        echo "STARTING HEAD at $node"
        ray start --head --node-ip-address=$ip --port=$port --redis-password=$redis_password --block &
        sleep 5
    else
        echo "STARTING WORKER at $node"
        ssh ${node} ray start --address=$ip_head --redis-password=$redis_password
        sleep 10
    fi
done

python do.py > out
```

stepan-tsirkin:
Nice! @jaemolihm , would you like to add a commit on that?  

You may add the snbatch template as another variable in __sbatch_template.py, and add another option to the slurm.py, and rename it in something more general 

Yes, the option ray-slurm should be renamed into ray -cluster.

jaemolihm:
I added PBS support. I renamed `slurm.py` to `cluster.py`, and added `--batch-system` command line argument.

Also, I changed the sleep times for tests (123.45 and 67.0) to 12.34 and 5.0, because it could give wrong impression to the users that such a long sleep time is required. If such long sleep is indeed needed, I will revert.

jaemolihm:
The `__cluster_template.py` file just contains two looong strings, `slurm_text` and `pbs_torque_text`. It could just go inside `cluster.py`, but it's a matter of taste.

jaemolihm:
I cleaned the cluster script test using `@pytest.mark.parametrize`: https://docs.pytest.org/en/6.2.x/parametrize.html.

Parametrization of tests is useful when the same test is run for different arguments. It could also be useful for `use_wcc_phase`.

stepan-tsirkin:
@jaemolihm , @Liu-Xiaoxiong  what do you think? To me looks like it is ready to merge. Any comments?

jaemolihm:
Sorry for the late comment. I think the branch is ready to merge.

Only one question: why was the test with python3.6 broken? If 3.6 is not supported (by ray or by wannierberri), it would be better to also update `python_requires` in `setup.py`.

jaemolihm (approved):

stepan-tsirkin:
> Only one question: why was the test with python3.6 broken? If 3.6 is not supported (by ray or by wannierberri), it would be better to also update python_requires in setup.py

Actually, it did pass for 3.6, However having 3 versions under testing made them to hang up randomly. I.e.it could be that 3.6 and 3.7 pass, but 3.8 hangs. Ot another one. I do not know if we can solve it other way, it looks like a bug/overload of github's testing server

jaemolihm:
It's strange.. Anyway, we can merge this PR and open an issue about the test.

PR #53, opened 2021-05-14
mopdified tests for AHC, added for the berry_dipole and other stuff,…

stepan-tsirkin:
… also for GaAs

I did not include "berry_dipole_fsurf, because for that I achieved ony 1e-5 (abs diff) agreement between wcc_phases and without that. Not sure if that is normal, or should we take more care for nonabelian stuff.


tests/conftest.py
```diff
@@ -9,3 +9,8 @@
 def rootdir():
     return os.path.dirname(os.path.abspath(__file__))
 
+
+@pytest.fixture(scope="session")
+def Efermi_Fe():
+    return os.path.dirname(os.path.abspath(__file__))
```
jaemolihm, line 15: Does this belong here?

stepan-tsirkin: For sure not

stepan-tsirkin:
@jaemolihm , do you understand why codecov thinks that line 77 is not covered by tests, while line 81 is covered?

Screenshot from 2021-05-14 10-37-09

jaemolihm:
Maybe because the two functions have the same name, so the first one is not run.

stepan-tsirkin:
> Maybe because the two functions have the same name, so the first one is not run.

Right. :+1:

stepan-tsirkin:
Interesting, on my computer tests work

stepan-tsirkin:
@jaemolihm , @xxliu could at least one of you have a fresh look and make a approving review.  Thanks.

Liu-Xiaoxiong:
Looks good.  It that degen_thresh help you approve E-5 to E-6? 

I will add "systems" here later

PR #55, opened 2021-05-14
Test system_tb and system_tb_py

Liu-Xiaoxiong:
Recording to conductivity ohmic.  Because the value is large (1e5,1e6). Sometimes the random error may larger than -e5 if you are lucky. So I set it -e4.

to test system_tbmodels and systems_pythtb. I have to install this two software. I put them in setup.py. I know it is not good. Any suggestion?

And if I run pytest on my local. I always have this bug. 
    
>           nb_out_list = wberri.mmn2uHu.run_mmn2uHu(seedname, INPUTDIR=data_dir,
                OUTDIR=str(data_dir)+"/reduced", **kwargs)
E           AttributeError: module 'wannierberri.mmn2uHu' has no attribute 'run_mmn2uHu'

Do you know how to fix it? I already update wannierberri to the newest version.


jaemolihm:
> to test system_tbmodels and systems_pythtb. I have to install this two software. I put them in setup.py. I know it is not good. Any suggestion?

You can add `pip install pythtb` or something similar in `.github/workflows/ci.yml`: see https://github.com/wannier-berri/wannier-berri/blob/master/.github/workflows/ci.yml#L25.

stepan-tsirkin:
> Sometimes the random error may larger than -e5 if you are lucky. So I set it -e4.

I think it is because conductivity_ohmic is large by itself, it is ~1e9 . Partially it is because it is evaluated at a large relaxation time of 1ns

Shopuldn't we better check the relative difference? but relative to what? I suggest to make it "relative to mean absolute value of an array" . could be also relative to the maximum value, but this is undesirable, if we have  a huge peak, everything becomes negligible compared to it.

Liu-Xiaoxiong:
> > to test system_tbmodels and systems_pythtb. I have to install this two software. I put them in setup.py. I know it is not good. Any suggestion?
> 
> You can add `pip install pythtb` or something similar in `.github/workflows/ci.yml`: see https://github.com/wannier-berri/wannier-berri/blob/master/.github/workflows/ci.yml#L25.

Great thanks.

stepan-tsirkin:
Why do you need to store the pickled systems?  they may be generated by a python script in a fixture. his will also add coverage of the system constructors

Liu-Xiaoxiong:
> > Sometimes the random error may larger than -e5 if you are lucky. So I set it -e4.
> 
> I think it is because conductivity_ohmic is large by itself, it is ~1e9 . Partially it is because it is evaluated at a large relaxation time of 1ns
> 
> Shopuldn't we better check the relative difference? but relative to what? I suggest to make it "relative to mean absolute value of an array" . could be also relative to the maximum value, but this is undesirable, if we have a huge peak, everything becomes negligible compared to it.

Yes, we can use the magnitude of the average -10 or -12.

stepan-tsirkin:
btw, how did you construct those models?

Liu-Xiaoxiong:
> Why do you need to store the pickled systems? they may be generated by a python script in a fixture. his will also add coverage of the system constructors

It is not a pickled system just pickled model of pythtb and tbmodels.

One reason is run pythtb and tbmodels in the test is really slow. (7 minutes more) 
Another reason is pickle file is much smaller than _hr.dat  wsvc._data.

We are testing WB not pythtb and tbmodesl.

Liu-Xiaoxiong:
> btw, how did you construct those models?

model_tbmodels = tbmodels.Model.from_wannier_files(
            hr_file='./Te_hr.dat',
            wsvec_file='./Te_wsvec.dat',
            xyz_file='./Te_centres.xyz',
            win_file='./Te.win'
                        )
Te =w90("./","Te")
model_pythtb=Te.model(min_hopping_norm=0.01)

Then pickle the models.

jaemolihm:
>  E AttributeError: module 'wannierberri.mmn2uHu' has no attribute 'run_mmn2uHu'

It runs fine on my laptop. Maybe something is wrong with the installation paths. Could you add `print(wberri.mmn2uHu.__file__)` before the line which errors and run `pytest -s` to see whether the installation path is for the recent version of wannier-berri? Note that you need to remove the `*.uHu` or other files if you already created them to rerun `mmn2uHu`.

stepan-tsirkin:
I see about the models. I mean, Maybe you can just add a toy model, like those in the tutorials Li and/or haldane.  It is fast and makes the test well.

It is true, we are testing WB and its interface to the TB model packages. So no need to use the conversion of wannier model to their systems.

stepan-tsirkin:
And first, merge the master branch and resolve conflicts

Liu-Xiaoxiong:
> And first, merge the master branch and resolve conflicts

Sure! After moving the installation of the two software to workflows! It's my time to sleep. Nice to see you at dawn.

Liu-Xiaoxiong:
Solved merge conflict. 
Auto_precisions =  1-E12 * (average of data_smear_data).
 If auto_precisions smaller than 1-E11. Set it to 1-E11.

tests/compare_result.py
```diff
@@ -41,7 +41,10 @@ def _inner(fout_name, suffix, adpt_num_iter,suffix_ref=None,precision=1E-10):
             E_titles, data_energy, data, data_smooth = read_energyresult_dat(filename)
             path_filename_ref = os.path.join(rootdir, 'reference', filename_ref)
             E_titles_ref, data_energy_ref, data_ref, data_smooth_ref = read_energyresult_dat(path_filename_ref)
-
+            if precision == None:
+                precision = abs(sum(sum(data_smooth_ref))/float(np.prod(np.shape(data_smooth_ref))*1E12) )
```
jaemolihm, line 45: Can be written as `np.average(abs(data_smooth_ref)) / 1E12`.

Also, is there a reason to use `data_smooth_ref` instead of `data_ref`?

jaemolihm:
I am against storing pickled system object inside the repository. In the future, we may change the `System_TBmodels` class. Then, the pickled data cannot be read and used correctly. Every time we change the class, we would need to regenerate the pickle file. (See [Alex Gaynor: Pickles are for Delis, not Software - PyCon 2014](http://www.youtube.com/watch?v=7KnfGDajDQw&t=21m32s) 21:32 ~ 24:30)

I think it is better to construct a toy tight-binding model on the fly like in the tutorials, as Stepan said.

Also, the `Fe_tb.dat.tar.gz` file takes a large disk space (~10MB). Is it needed? Just removing it in a commit will still increase the repository size because it stays in git history. But I am not aware of an easy solution. (A hard-and-dirty solution is to find the commit where the large file is added, rebase the commit history to edit that commit, and open another PR.)

Liu-Xiaoxiong:
> I am against storing pickled system object inside the repository. In the future, we may modify the `System_TBmodels` class (e.g. rename attributes or add new attributes). Then, the pickled data cannot be read and used correctly. Every time we change the class, we would need to regenerate the pickle file. (See [Alex Gaynor: Pickles are for Delis, not Software - PyCon 2014](http://www.youtube.com/watch?v=7KnfGDajDQw&t=21m32s) 21:32 ~ 24:30)
> 
> I think it is better to construct a toy tight-binding model on the fly like in the tutorials, as Stepan said.
> 
> Also, the `Fe_tb.dat.tar.gz` file takes a large disk space (~10MB). Is it needed? Just removing it in the next commit will still increase the repository size because the file stays in git history. But I am not aware of an easy solution. (A hard-and-dirty solution is to find the commit where the large file is added, rebase the commit history to edit that commit, and open another PR.)

Sure, I will change to a toy model. 

Regarding 'Fe_tb.dat.tar.gz'. System_tb needs a _tb.dat as the input file. I agree _tb file is large. I can generate a new one with less DFT grid. It will help to reduce the size of the file.

Liu-Xiaoxiong:
> I am against storing pickled system object inside the repository. In the future, we may modify the `System_TBmodels` class (e.g. rename attributes or add new attributes). Then, the pickled data cannot be read and used correctly. Every time we change the class, we would need to regenerate the pickle file. (See [Alex Gaynor: Pickles are for Delis, not Software - PyCon 2014](http://www.youtube.com/watch?v=7KnfGDajDQw&t=21m32s) 21:32 ~ 24:30)
> 
> I think it is better to construct a toy tight-binding model on the fly like in the tutorials, as Stepan said.
> 
> Also, the `Fe_tb.dat.tar.gz` file takes a large disk space (~10MB). Is it needed? Just removing it in the next commit will still increase the repository size because the file stays in git history. But I am not aware of an easy solution. (A hard-and-dirty solution is to find the commit where the large file is added, rebase the commit history to edit that commit, and open another PR.)

Does it help if we cancel this branch, and build a new branch based on the master branch? And open a new PR?

tests/compare_result.py
Liu-Xiaoxiong: Thanks!

Using data_smooth_ref because I am afraid the peaks from no_smooth data will interfere with the estimate of the average.

jaemolihm:
> Regarding 'Fe_tb.dat.tar.gz'. System_tb needs a _tb.dat as the input file. I agree _tb file is large. I can generate a new one with less DFT grid. It will help to reduce the size of the file.

I see, thanks for the explanation. So we need at least one `_tb.dat` file. You could use the existing GaAs data instead of the Fe data, as the GaAs one uses less WFs and a smaller grid.

> Does it help if we cancel this branch, and build a new branch based on the master branch? And open a new PR?

Yes, it will help.

PR #56, opened 2021-05-15
Tests: Write test output to a single folder, use higher smearEf

jaemolihm:
* All `.dat` files created by the test are written in a single temporary folder `tests/_dat_files`.
* Use higher smearEf value (600K) for berry tests. The default value (10K) is too small to make visible changes.


stepan-tsirkin (approved):
Looks good to me

PR #57, opened 2021-05-19
Reorganize Smoother

jaemolihm:
I reorganized the Smoother modules. The purpose is to make implementation of new Smoothers easier.

I defined an AbstractSmoother, which defines `smt`, `__call__`, and a few other methods, but not `_broaden`. Each specific Smoother inherits the AbstractSmoother and only implements its own `_broaden`. VoidSmoother is a special case in which `__call__` is also overriden.

I renamed `Smoother` to `FermiDiracSmoother` and added `GaussianSmoother`. Currently, `GaussianSmoother` is not used in the code, but later it might be possible to use it to smooth functions of frequencies, in e.g. Kubo routines. (But it needs some thinking because for frequencies, in the zero-smearing limit, we have delta functions, not step functions as for occupations.)

Also, I moved some LazyProperties to `__init__` because they were just a few real numbers.

### DIFF
--- a/tests/compare_result.py
+++ b/tests/compare_result.py
@@ -48,7 +48,7 @@ def compare_energyresult(output_dir, rootdir):
             E_titles_ref, data_energy_ref, data_ref, data_smooth_ref = read_energyresult_dat(path_filename_ref)
 
             assert E_titles == E_titles_ref
-            # assert data_energy == approx(data_energy_ref, abs=precision)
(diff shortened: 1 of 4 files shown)


stepan-tsirkin (approved):
Looks fine.

stepan-tsirkin:
@jaemolihm , I think now, after approval, you can merge yourself (right, you see the button green?).

PR #59, opened 2021-05-23
tests for symmetry. renamed test_berry.py-> test_integrate.py. 

stepan-tsirkin:
Also, because refinement tests are unstable, they are performed only for selected calculations, and only with symmetries. In particular due to numerical instabilities, a different instance of the symmetry-equivalent k-points may be chosen. Also refinement depends on smearing and on addition of extra quantities to calculate.


tests/test_integrate.py
```diff
@@ -84,6 +91,11 @@ def test_Fe_wcc(check_integrate,system_Fe_W90_wcc, compare_energyresult,quantiti
     check_integrate(system_Fe_W90_wcc , quantities_Fe , fout_name="berry_Fe_W90" , suffix="wcc" , Efermi=Efermi_Fe , comparer=compare_energyresult )
 
 
+def test_Fe_sym(check_integrate,system_Fe_W90_sym, compare_energyresult,quantities_Fe,Efermi_Fe):
+    """Test anomalous Hall conductivity , ohmic conductivity, dos, cumdos"""
+    check_integrate(system_Fe_W90_sym , quantities_Fe , fout_name="berry_Fe_W90" , suffix="sym" , suffix_ref="sym", Efermi=Efermi_Fe , comparer=compare_energyresult )
```
jaemolihm, line 96: What about using `fout_name="berry_Fe_W90-sym"` and removing `suffix` and `suffix_ref` arguments?

stepan-tsirkin: good point. Actually, for tight-binding we can use 
```
fout_name="berry_Fe_W90" , suffix="sym" , suffix_ref=""
```
(implemented in a new commit). because for them the symmetrized results are the same. And we may use the old references. However, with refinement we need to use new references, because the refinement works differently with symetries.

So, it is better to keep here in the same style.
I think, it may be eventually useful to add more symmetric tb models, to test more of the symmetry operations.

jaemolihm (approved):

PR #64, opened 2021-06-01
version 1.0.0

stepan-tsirkin:
I created this branch to put here commits that modify the API.  

fixes that do not require change of API can be pushed to the master branch.


wannierberri/__Data_K.py
```diff
@@ -28,59 +28,87 @@ def _rotate_matrix(X):
...
+        return self.system.CC_R*self.expdK[:,:,:,None]
+
+    @lazy_property.LazyProperty
+    def SS_R(self):
+        return self.system.SS_R*self.expdK[:,:,:,None]
+
+    @property
+    def SH_R(self):
+        return self.system.SH_R*self.expdK[:,:,:,None]
+
```
jaemolihm, line 134: Why is SH_R (and all other spin Hall related matrices) a `@property` while others are `@LazyProperty`?

stepan-tsirkin: because they are used only once in the class, and the use is in another lazy_property `SA_H`, `SHA_H` or `shc_B_H` so, there is no need to store `SH_R`. Also removed corresponding  copy() for that reason.  `AA_R` , `BB_R`, `CC_R` are used at least twice.

stepan-tsirkin:
@jaemolihm @Liu-Xiaoxiong @manxkim 

I suggest to merge now to mastrer, after you review the code. The branch has deviated so much, that there is no point in maintaining the old master branch. Just in case, I branched the master int [this](https://github.com/wannier-berri/wannier-berri/tree/branch-0.9.x), just in case we want to make a minor fix. 

This does not mean that we release the version 1-0-0, but that we are on a straight line before that. I think we can announce the alpha soon, after we agree on the naming and structure of principle classes, modules and input parameters. THerefore, please pay a close attention to this. 

Keeping this branch separate already makes problems for further development, so let's not postpone the merge too much. 

I also created a corresponding branch in the [documentation](https://github.com/wannier-berri/wannier-berri-org/tree/version-1.0.0) , please look at it and contribute what you can.  This version of documentation, being under construction, is also uploaded to the [server ](https://wannier-berri.org/version-1-0-0/)

Let's do it!

jaemolihm:
Great! Thank you a lot for the effort. We are now very close to v1.0!

Could you first resolve the confilcts by merging the current master to this branch? Also, to avoid having more conflicts, I suggest to pause adding new features to master until we merge this branch.

Also for the ease of the review, could you summarize the changes and new features? For example, I see lots of reference data for tests are modified.

Ideally, it may be very helpful if we could split this giant PR into several PRs with one specific goal in each (e.g. new tests, new feature A, B, C, deprecated feature a, b, c, ...).

stepan-tsirkin:
>Could you first resolve the confilcts by merging the current master to this branch? 

Done!

>Also, to avoid having more conflicts, I suggest to pause adding new features to master until we merge this branch.

Agree, let's not merge anything anymore. And not add anything NEW to this branch, except necessary fixes.

> Also for the ease of the review, could you summarize the changes and new features? For example, I see lots of reference data for tests are modified.

I started summarizing here : https://wannier-berri.org/version-1-0-0/newfeatures.html

The principle things   are:

-  we switched to the new way of handling fermi-level scan - FermiOcean (it was v3 already, but '3' was removed from the naming of files/modules yesterday)
- some reference data had to be changed, in particular for Fermisurface quantities ('nonabelian' in old version) because of a different handling of the scan. In the old version, a band was contributing to the closest Fermi level (above or below its energy) in the list [here](https://github.com/wannier-berri/wannier-berri/blob/a269e5fc744f4ae442f4e9780ed7c245151369a3/wannierberri/__nonabelian.py#L275-L283)  But in new version we actually take a [derivative](https://github.com/wannier-berri/wannier-berri/blob/c66d9d4669a589d1a0c31198d81e0089394ebd74/wannierberri/fermiocean.py#L282-L291) of the fermi-dirac distribution , so in case `fder=1` a band contributes both to Fermi level below and above, but with factors 1/2. In the dense-EF limit  (when Fermi-level spacing is smaller than the smearing) this is the same - in the smeared data. But nonsmeared are different, and withsmall number of Fermi levels we get a big deiiference in the tests
- Also we discovered a bug in orbital magnetization, therefore a change
- tetrahedron method is implemented and new tests are there
- evaluation is based on "covariant formulae" . This allows to implement basic ingredients, and use them for different quantiuties. This makes implementation of new features more robust (less room for a mistake). Also tabulation uses the same formulae
- reorganized input parameters please have a look, is it convenient, is the naming  `specific_parameters` intuitive?
- now both tabulating and integratingg may be done with user-defined quantities. (no tests so far - to be done)

>Ideally, it may be very helpful if we could split this giant PR into several PRs with one specific goal in each (e.g. new tests, new feature A, B, C, deprecated feature a, b, c, ...).

I am afraid, that would be a hard work to do, which would produce many more bugs. Let's wrk on it as a whole. We did quite extensive testing with @Liu-Xiaoxiong ,and quite shure that this branch gives correct results. (or even corrected the older bugs). It is now more robust, because leaves less room to make  mistake in one thing, but not the other.

tests/compare_result.py
```diff
@@ -50,9 +50,9 @@ def _inner(fout_name, suffix, adpt_num_iter,suffix_ref=None,precision=None,compa
             E_titles_ref, data_energy_ref, data_ref, data_smooth_ref = read_energyresult_dat(path_filename_ref)
 
             if precision is None:
-                precision = max(abs(np.average(data_smooth_ref) / 1E12), 1E-11)
+                precision = max(abs(np.average(data_smooth_ref) / 1E11), 1E-10)
```
jaemolihm, line 53: Why loosen the test tolerance?

tests/test_tabulate.py
```diff
@@ -0,0 +1,120 @@
...
+#    compare= {'ahc_ocean':'ahc','ahc3_ocean':'ahc',"cumdos3_ocean":"cumdos","dos3_ocean":"dos","berry_dipole_ocean":"berry_dipole","berry_dipole3_ocean":"berry_dipole",
+#            'conductivity_ohmic3_ocean':'conductivity_ohmic','conductivity_ohmic_fsurf3_ocean':'conductivity_ohmic_fsurf'}
+    compare = {}
+    if quant in compare:
+        return compare[quant]
+    else:
+        return quant
+
+
+def test_Fe(check_tabulate,system_Fe_W90, compare_fermisurfer,quantities_tab):
```
jaemolihm, line 90: Is any of these bands degenerate? (on some high symmetry k point). If so, there may be a risk that some quantities like band velocity might be different between runs,

wannierberri/__integrate.py
```diff
@@ -130,15 +138,20 @@
...
-key='kpart'
-for calc in calculators_trans:
-    if calc.endswith('_ocean'):
-        additional_parameters[calc][key] = 500
-        additional_parameters_description[calc][key] = (
-             'Separate k-points of the FFT grid into portions ' + 
-             '(analog of ksep in the system class, but acts in different calculators)'  +
-             'decreasing this parameter helps to save memory in some cases' +
-                'while performance is usually unafected' )
+#for calc in calculators_trans:
```
jaemolihm, line 141: Please Delete these comments if they are not relevant any more. If not, please add some comment on why there are commented out but still there.

wannierberri/__main.py
```diff
@@ -133,16 +136,32 @@ def integrate(system,grid,Efermi=None,omega=None, Ef0=0,
...
         number of recursive adaptive refinement iterations. See :ref:`sec-refine`
     adpt_fac : int 
         number of K-points to be refined per quantity and criteria.
     parallel : :class:`~wannierberri.Parallel`
         object describing parallelization scheme
-    numproc : int 
-        (obsolete, use parallel instead) number of parallel processes. If <=0  - serial execution without `multiprocessing` module.
-    chunksize : int
-        (obsolete, use parallel instead) chunksize for distributing K points among processes. If not set or if <=0, set to max(1, min(int(numK / num_proc / 200), 10)). Relevant only if num_proc > 0.
+    irkpt : bool
```
jaemolihm, line 152: Irkpt is a bit too cryptic to me… maybe irred_kpt? use_irr_kpt? (Like use_wcc_phase) Irreducible_kpt?

wannierberri/__system.py
```diff
@@ -73,90 +68,22 @@ class System():
...
-        print ("Number of wannier functions:",self.num_wann)
-        print ("Number of R points:", self.nRvec)
-        print ("Recommended size of FFT grid", self.NKFFT_recommended)
-        print ("Real-space lattice:\n",self.real_lattice)
-        #print ("R - points and dege=neracies:\n",iRvec)
-        has_ws=str2bool(f.readline().split("=")[1].strip())
+    Notes:
+    -------
+        for tight-binding models it is recommended to use `use_wcc_phase = True`. In this case the external terms vanish, and 
+        one can safely use `berry=False, morb=False`, and also set `'external_terms':True` in the parameters of the calculation
```
jaemolihm, line 83: Should this be external terms : false? (In the sense that external terms can be neglected)

wannierberri/__tetrahedron.py
```diff
@@ -0,0 +1,304 @@
...
+                    bandmax=min(bandmax, bands_in_range[0][0])
+                weights[(0,bandmax)]=self.ones
+#                print ("now bandmax=",bandmax)
+
+            res.append( weights )
+        return res
+
+
+
+if __name__ == '__main__' :
```
jaemolihm, line 246: Should his part be here? Maybe move to tutorial or tests?

jaemolihm (changes requested):
I skimmed through the changes. In addition to the comments, I suggest some general things to do:

- [ ] check how much the repository size will increase
- [ ] Make sure that all changes in the test reference data are understood and *documented* (as a comment in this PR). (E.g. “change in the Fermi level assignment for bands: test data A, B, C, …, “morb bug fixed”: test data D, E)

I noticed that some parts are copied from an old version of the code, but we can just merge and deal with those later.

tests/test_tabulate.py
stepan-tsirkin: This should not happen, because there is a degeneracy threshold (1e-4), and for degenerate bands the values are averaged. This avoids gauge dependence.

jaemolihm: Okay, great!

jaemolihm:
I checked that the repository size does not increase much:
```bash
jmlihm@DESKTOP-K0F0FN5:~/temp$ git clone https://github.com/wannier-berri/wannier-berri.git --single-branch --branch master
jmlihm@DESKTOP-K0F0FN5:~/temp$ du -sh wannier-berri/
125M    wannier-berri/

jmlihm@DESKTOP-K0F0FN5:~/temp$ rm -rf wannier-berri/
jmlihm@DESKTOP-K0F0FN5:~/temp$ git clone https://github.com/wannier-berri/wannier-berri.git --single-branch --branch version-1.0.0
jmlihm@DESKTOP-K0F0FN5:~/temp$ du -sh wannier-berri/
127M    wannier-berri/     
```

wannierberri/__tetrahedron.py
stepan-tsirkin: Done, moved to tests

wannierberri/__integrate.py
stepan-tsirkin: Done, removed many commented lines in the code

wannierberri/__system.py
stepan-tsirkin: right, corrected

tests/compare_result.py
stepan-tsirkin: right, no need for that

stepan-tsirkin:
> * Make sure that all changes in the test reference data are understood and _documented_ (as a comment in this PR). (E.g. “change in the Fermi level assignment for bands: test data A, B, C, …, “morb bug fixed”: test data D, E)

@jaemolihm @Liu-Xiaoxiong @manxkim 

I collected the draft of the PR comment in this [file](https://github.com/wannier-berri/wannier-berri/blob/version-1.0.0/comment-PR-version-1.0.0.rst) .  If you feel that something else should be mentioned there, do not hesitate to contribute to this file. 

@jaemolihm , is there anything that I missed from your comments?  I think I responded to all of them and made proper corrections. 

If I do not here strong objections from you before Friday, I will merge on Friday (Nov 05). Or even earlier, if I get 2 approving reviews from you. 

if you see minor issues, you can add your own PRs later to correct them.

jaemolihm (approved):

jaemolihm:
Thanks a lot for the work! Yes, all my comments are addressed. The documentation is also very clear.

I added deprecation of `multiprocessing` to the PR comment file.

PR #65, opened 2021-06-14
Add shift current tests, refactor test_kubo.py

jaemolihm:
I added shift current tests and refactored `test_kubo.py`.

I tried to implement `use_wcc_phase=True` for SHC without success, so I left it as a TODO.


stepan-tsirkin (approved):
Looks fine

PR #70, opened 2021-07-16
Bugfix and small optimization for shift current

jaemolihm:
* Bugfix for shift current (the frequency-dependent `delta` factor was wrongly overwritten for multiple Fermi levels, so that the results were wrong starting from the second Fermi level.)
* reference data for tests updated
* Optimization for multiple Fermi levels (moving `delta` calculation outside EFermi loop)


stepan-tsirkin (approved):
Makes total sense

PR #71, opened 2021-08-16
Ray fix for error on shutdown

jaemolihm:
With the new Parallel class, using ray was giving error when shutting down python (can be reproduced by running example.py in tutorial). The reason is that `__del__` is called when shutting down python, but imported modules are already deleted, so python cannot run `ray.shutdown()`.

This error may be related to https://github.com/wannier-berri/wannier-berri/pull/52#issuecomment-886157263.

Here, I renamed `__del__` to `shutdown` so that it is not called when shutting down python. Users can close their Parallel object by `parallel.shutdown()`. I also added python 3.6 test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     timeout-minutes: 20
     strategy:
       matrix:
-        python-version: [ 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8]
         install-type: [dev]
 
     steps:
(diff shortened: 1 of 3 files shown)


stepan-tsirkin:
Thank you @jaemolihm  . 

One note :  although it is not necessary, could you add `parallel.shutdown()` to the [example](https://github.com/jaemolihm/wannier-berri/blob/ray_fix/tutorial/parallel-ray/example.py) , for illustrative purposes.

jaemolihm:
Done.

stepan-tsirkin (approved):
All fine

PR #72, opened 2021-08-22
Various fixes

jaemolihm:
1. Fix a bug that `mmn.data` is changed when computing `BB`. This happens when `morb` and `SHCqiao` are both true.

2. Add some comments on `parallel_ray` fixture. If in the future a `Parallel` object with a different ray setting is included in the test suite, one needs to take care to shutdown parallel object before initiating a new one.

3. Make `rootdir` and `output_dir` constant functions. Since their values are not changed, they need not be fixtures. I think same holds for the various parameters in `test_integrate.py`: they can just be constants.
The benefit is that they can be used in non-fixture functions.

### DIFF
--- a/tests/compare_result.py
+++ b/tests/compare_result.py
@@ -6,7 +6,7 @@ import numpy as np
 import pytest
 from pytest import approx
 
-from conftest import rootdir, output_dir
+from conftest import ROOT_DIR, OUTPUT_DIR
 
 def read_energyresult_dat(filename):
     """Read .dat file output of EnergyResult."""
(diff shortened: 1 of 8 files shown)


stepan-tsirkin:
Thak you, @jaemolihm , well spotted `mmn.data` 

One question : what is the motivation of using constant functions vs fixtures, and why not using here "constant" values instead? Is it because functions are harder to re-define accidentally ? Is there some common guidance on this point?

stepan-tsirkin (approved):
All fine for me, you may merge

jaemolihm:
> One question : what is the motivation of using constant functions vs fixtures, and why not using here "constant" values instead? Is it because functions are harder to re-define accidentally ? Is there some common guidance on this point?

Good point. I (wrongly) thought that constant values will not work, but they do work. I used constant values, and changed `rootdir` and `output_dir` to `ROOT_DIR` and `OUTPUT_DIR` to be consistent with the python style guide (https://www.python.org/dev/peps/pep-0008/#constants).

Since there are no true constants (that errors when the value is reassigned) in python, constant functions could be used to make sure that the value does not change. But I think when it is clear that the values will not be reassigned, constant values are simpler to use (no `()`).

The motivation for using constant values instead of fixtures is that 1) fixtures need to be listed in the arguments of each test function, and 2) fixtures cannot be used inside functions which are not fixtures.

PR #73, opened 2021-09-13
Set smearW=None as default, fix typos

jaemolihm:
* In `integrate`, `smearW` is imposed to be `None` (i.e. not used), but the default was `10`. I changed the default to `None`, and updated the warning that is printed if it is not `None`.
* Fixed typos: "qantities" -> "quantities", "%%" -> "%"

### DIFF
--- a/wannierberri/__grid.py
+++ b/wannierberri/__grid.py
@@ -98,7 +98,7 @@ class Grid():
                                 K_list[k[0]][k[1]][k[2]]=None
         K_list=[K for Kyz in K_list for Kz in Kyz for K in Kz if K is not None]
         print ("Done in {} s ".format(time()-t0))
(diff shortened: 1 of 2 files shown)


stepan-tsirkin:
Hi @jaemolihm  , 
if parameter smearW is not used, why don't we just remove it, to avoid confusion?

And then just set smootherOmega=VoidSmoother()
 https://github.com/wannier-berri/wannier-berri/pull/73/files#diff-9ecc5a5e191efe0be2728e811700f7c2e618a0cbcd41c87aa245c9c8afb2004eR183

jaemolihm:
Hi @stepan-tsirkin , currently, we can just remove it. But in the future, I think `smearW` will be useful to introduce the optimization that is done to Fermi level scan to frequency scan. So I propose to let it stay there until we implement frequency scan optimization.
`smearW` was already removed from the function docstring, so I think there will not be much confusion from the user side.

stepan-tsirkin (approved):
Ok, let it be so far

PR #74, opened 2021-09-13
pickle Klist by seperating Kpoints to save memory usage

Liu-Xiaoxiong:


wannierberri/__evaluate.py
```diff
@@ -154,7 +154,14 @@ def paralfunc(Kpoint,_system,_grid,npar_k,fftlib):
...
     if restart:
         try:
-            K_list=pickle.load(open(file_Klist,"rb"))
+            fr = open(file_Klist,"rb")
+            K_list = []
+            
+            while True:
+                try:
+                    K_list +=pickle.load(fr)
+                except:
```
stepan-tsirkin, line 163: Maybe specify, what exception to expect. `EOFError` ?

wannierberri/__evaluate.py
```diff
@@ -203,7 +210,14 @@ def paralfunc(Kpoint,_system,_grid,npar_k,fftlib):
...
             if file_Klist is not None:
-                pickle.dump(K_list,open(file_Klist,"wb"))
+                
+                nk = len(K_list)
+                fw =  open(file_Klist,"wb")
+                nkpart = nk//10
+                for ink in range(nkpart):
+                    pickle.dump(K_list[10*ink:10*(ink+1)],fw)
+                if nk%10 !=0:
+                    pickle.dump(K_list[10*nkpart:],fw)
```
stepan-tsirkin, line 220: it can be done in one `for` loop, e.g. 
```
for i in range(0,nk,10):
    pickle.dump(K_list[i:i+10],fw)
```

wannierberri/__evaluate.py
```diff
@@ -203,7 +210,14 @@ def paralfunc(Kpoint,_system,_grid,npar_k,fftlib):
         
         try:
             if file_Klist is not None:
-                pickle.dump(K_list,open(file_Klist,"wb"))
+                
+                nk = len(K_list)
+                fw =  open(file_Klist,"wb")
+                nkpart = nk//10
```
stepan-tsirkin, line 216: Why exactly 10 ? could you make it a free parameter?

stepan-tsirkin (changes requested):
Could you also add a test . Maybe pickle after the 0th iteration, and then restart and do the 1st iteration, and compare with the existing data for the test https://github.com/wannier-berri/wannier-berri/blob/9da92db168568bdad63cc21d05005603f7c0214e/tests/test_integrate.py#L156-L160

Should be the same, right?

wannierberri/__evaluate.py
Liu-Xiaoxiong: except here is working to break the while loop after reading Klist

wannierberri/__evaluate.py
Liu-Xiaoxiong: I don't think so, nk%10 is not always 0.

wannierberri/__evaluate.py
stepan-tsirkin: Yes, but there may be different exceptions. Say, if it is EOFError, then it is fine, it is just means 
but if there is another error (say, the object cannot be unpickled, because it was pickled because it is corrupt, I do not know why), than the code should raise it and exit.

wannierberri/__evaluate.py
stepan-tsirkin: Not a problem 
Screenshot from 2021-09-14 20-49-37

wannierberri/__evaluate.py
stepan-tsirkin: Screenshot from 2021-09-14 20-56-06

stepan-tsirkin: Screenshot from 2021-09-14 20-57-43

stepan-tsirkin (approved):

Issue #63, opened 2021-05-29
use_wcc_phase for orbital magnetization and spin Hall


jaemolihm:
#45 introduced the use_wcc_phase option, which enables Berry curvature dipole (and some other) calculation with the so-called "tight-binding convention" (`use_wcc_phase=True`).

I tested spin Hall conductivity, and `use_wcc_phase=True` gave different results compared to `use_wcc_phase=False`, probably because <0m|Sr|Rn> and <0m|SHr|Rn> need some special treatment (their own `flag` in `_R_to_k_H`).
I guess similar problem exists in the orbital magnetization (but I have not tested this), because BB_R (<0m|H(r-R)|Rn>) and CC_R (<0m|rH(r-R)|Rn>) need special treatment (their own `flag`). Is this correct?

There are several `flag`s in `Data_K`, e.g. `flag='VV'`, but only `flag='AA'` is implemented. If they are not needed, we should delete them. If they are needed but not implemented, we should raise a `NotImplementedError` in `_R_to_k_H` until we get it fixed.

Also, I wonder what `flag` does can be done at the `system` level. (I remember this was discussed somewhere in #45.) If this can be done, tight-binding calculations can be faster because `AA_R` would be zero. @Liu-Xiaoxiong, are you working on this?

I'll try to implement `use_wcc_phase=True` for the spin Hall calculation.

@Liu-Xiaoxiong @stepan-tsirkin

stepan-tsirkin:
Yes, every matrix element should be treated carefully. Xiaoxiong is working on that for the orbital moment. 

One thing that I do not really like about use_wcc_phase, is that it is a parameter of the system. While it shopuld be a parameter of the calculation. So, to change the parameter, we change the system, which is not logical. Instead, we may make a list of general parameters of the calculation, and pass them to integrate() and tabulate( ) as a dictionary, which further may be passed to the Data_K.__init__(). 

Just, I want that a system can be initialized only once, then pickled (or JSON'ed, as you like) and then used for multiple calculations. In particular, the tests will run faster then.

This also applies to other parameters, like `transl_inv`, and some others.

For example, symmetries should be a property of the system, but to use them or not - should be a parameter of the calculation.

jaemolihm:
> Just, I want that a system can be initialized only once, then pickled (or JSON'ed, as you like) and then used for multiple calculations. In particular, the tests will run faster then.

[...]

PR #77, opened 2021-09-22
Improved wcc

stepan-tsirkin:
Changed implementation of use_wcc_phase. Now (almost) everything is done on the level of System()

Also implemented  use_wcc_phase for the Morb  calculation. However, it requires a correction term.  
 However, a notable difference is still observed with use_wcc_phase = False, not sure why.


jaemolihm:
I found that achieving phase convention independence is quite tricky. For example, the “Ryoo” formula for spin current operator is actually not convention independent. Maybe there is a similar problem in morb.
Anyway, there is an open issue for this (#63), so we can track the issue separately from this MR if you want.

wannierberri/__Data_K.py
```diff
@@ -933,13 +923,15 @@ def Hplusminus(self,sign,evalJ0=True,evalJ1=True,evalJ2=True):
...
-                res['ii']=-2*self.A_E_A
+                res['ii']+=-2*self.A_E_A
             res['i']+=self.Morb_Hbar_diag + sign*self.Omega_Hbar_E
         if evalJ1:
             res['oi']+=-2*(self.D_B+sign*self.D_E_A)
         if evalJ2:
             C,D=self.D_E_D
             res['oi']+=-2*(C+sign*D)
+        if self.use_wcc_phase:
+            res['ii']+=2*self.T_BmEA
```
stepan-tsirkin, line 934: This is a term to correct dependence on wcc_phase

stepan-tsirkin:
Yes, it is a tricky thing. With orbital moment I found that the difference between the two conventions is proportional to a term, which should vanish in principle, but comes out non-zero in Wannier interpolation. That is why I added a correction term, but for some reason it does not work out. I will share notes later, and maybe a similar trick is needed for SHC. 

 Maybe I messed smth, because the old implementation is not intuitive to me anymore. 

I still sugggest to merge it now, at least it improves the working functionalities without breaking it. The wcc_phase for Morb was not implemented before. Later I will merge to the development [branch](https://github.com/wannier-berri/wannier-berri/tree/formula+berry-boltzmann) and try to fix it there. 

@jaemolihm  do you agree?

jaemolihm:
> Yes, it is a tricky thing. With orbital moment I found that the difference between the two conventions is proportional to a term, which should vanish in principle, but comes out non-zero in Wannier interpolation. That is why I added a correction term, but for some reason it does not work out. I will share notes later, and maybe a similar trick is needed for SHC.

I agree. I will study the notes when you share them. I think an important question is what is the "true" answer. What I found in my Comment to your shift current paper (with Julen and Ivo) was that both conventions needed a (different) correction term to get the "true" answer (true in the sense that it is defined in the Kohn-Sham basis before Wannierization), although there the use of `sc_eta` was the issue.

> I still sugggest to merge it now, at least it improves the working functionalities without breaking it.

Yes, I agree to merge it now. I will do the review in a few days.

wannierberri/__system.py
```diff
@@ -284,23 +289,130 @@ def set_symmetry(self,symmetry_gen=[]):
...
+
+    @lazy_property.LazyProperty
+    def diff_wcc_red(self):
+        """ 
+        With self.use_wcc_phase=True it is tj-ti. With self.use_wcc_phase=False it is 0. [m,n,a] (Reduced)
+        """
+        wannier_centers = self.wannier_centers_reduced
+        return np.array([[j-i for j in wannier_centers] for i in wannier_centers])
+
+    def set_wannier_centers(self):
```
jaemolihm, line 325: 1. This function does two things. Setting wannier centers itself, and modifying matrix elements. Could you mention this in a comment?
2. Should we add a barrier to ensure this function is not called twice? Oh, it seems that in that case the ValueError will be raised. Maybe add there “, or, set_wannier_centers should not be called twice”?

wannierberri/__system.py
```diff
@@ -284,23 +289,130 @@ def set_symmetry(self,symmetry_gen=[]):
...
+#                    raise NotImplementedError(f"use_wcc_phases=True is not implemented for {X}_R")
+
+
+    @property
+    def iR0(self):
+        return self.iRvec.tolist().index([0,0,0])
+
+    @lazy_property.LazyProperty
+    def reverseR(self):
+        """maps the R vector -R"""
```
jaemolihm, line 386: R vector “to” -R? Or more precisely, it is the indices of R vectors that has -R in irvec, and the indices of the corresponding -R vectors.

wannierberri/__system.py
```diff
@@ -284,23 +289,130 @@ def set_symmetry(self,symmetry_gen=[]):
...
+        iRveclst= self.iRvec.tolist()
+        mapping = np.all( self.iRvec[:,None,:]+self.iRvec[None,:,:] == 0 , axis = 2 )
+        # check if some R-vectors do not have partners
+        notfound = np.where(np.logical_not(mapping.any(axis=1)))[0]
+        for ir in notfound:
+            print ("WARNING : R[{}] = {} does not have a -R partner".format(ir,self.iRvec[ir]) )
```
jaemolihm, line 392: Use f string?

wannierberri/__system.py
```diff
@@ -284,23 +289,130 @@ def set_symmetry(self,symmetry_gen=[]):
...
+        # check if some R-vectors have more then 1 partner 
+        morefound = np.where(np.sum(mapping,axis=1)>1)[0]
+        if len(morefound>0):
+            raise RuntimeError( "R vectors number {} have more then one negative partner : \n{} \n{}".format(
+                            morefound,self.iRvec[morefound],np.sum(mapping,axis=1) ) )
+        lst1,lst2=[],[]
```
jaemolihm, line 398: lst_R and lst_minusR(or mR)?

wannierberri/__system.py
```diff
@@ -284,23 +289,130 @@ def set_symmetry(self,symmetry_gen=[]):
...
+                lst1.append(ir1)
+                lst2.append(ir2[0])
+#                print (ir1,self.iRvec[ir1] , ir2,self.iRvec[ir2[0]])
+        return np.array(lst1),np.array(lst2)
+
+    def conj_XX_R(self,XX_R):
+        """ reverses the R-vector and takes the hermitian conjugate """
+        XX_R_new = np.zeros_like(XX_R)
+        lst1,lst2 = self.reverseR
+        assert np.all(self.iRvec[lst1] + self.iRvec[lst2] ==0 )
```
jaemolihm, line 411: I think his should go to reverseR

wannierberri/__system.py
```diff
@@ -284,23 +289,130 @@ def set_symmetry(self,symmetry_gen=[]):
...
+#        print (XX_R.shape,XX_R_new.shape,lst1,lst2)
+        XX_R_new [:,:,lst1] = np.copy(XX_R)[:,:,lst2]
+        XX_R_new[:] = XX_R_new.swapaxes(0,1).conj()
+        return np.copy(XX_R_new)
```
jaemolihm, line 415: Why return a copy?

wannierberri/__system.py
```diff
@@ -284,23 +289,130 @@ def set_symmetry(self,symmetry_gen=[]):
...
+        return np.array(lst1),np.array(lst2)
+
+    def conj_XX_R(self,XX_R):
+        """ reverses the R-vector and takes the hermitian conjugate """
+        XX_R_new = np.zeros_like(XX_R)
+        lst1,lst2 = self.reverseR
+        assert np.all(self.iRvec[lst1] + self.iRvec[lst2] ==0 )
+#        print (XX_R.shape,XX_R_new.shape,lst1,lst2)
+        XX_R_new [:,:,lst1] = np.copy(XX_R)[:,:,lst2]
+        XX_R_new[:] = XX_R_new.swapaxes(0,1).conj()
```
jaemolihm, line 414: Delete [:]?

wannierberri/__system.py
```diff
@@ -284,23 +289,130 @@ def set_symmetry(self,symmetry_gen=[]):
...
+#                print (ir1,self.iRvec[ir1] , ir2,self.iRvec[ir2[0]])
+        return np.array(lst1),np.array(lst2)
+
+    def conj_XX_R(self,XX_R):
+        """ reverses the R-vector and takes the hermitian conjugate """
+        XX_R_new = np.zeros_like(XX_R)
+        lst1,lst2 = self.reverseR
+        assert np.all(self.iRvec[lst1] + self.iRvec[lst2] ==0 )
+#        print (XX_R.shape,XX_R_new.shape,lst1,lst2)
+        XX_R_new [:,:,lst1] = np.copy(XX_R)[:,:,lst2]
```
jaemolihm, line 413: Dont need copy

wannierberri/__system_tb.py
```diff
@@ -72,23 +72,22 @@ def __init__(self,tb_file="wannier90_tb.dat",**parameters):
...
+            self.AA_R=np.zeros( (self.num_wann,self.num_wann,nRvec,3) ,dtype=complex)
+            for ir in range(nRvec):
+                f.readline()
+                assert (np.array(f.readline().split(),dtype=int)==self.iRvec[ir]).all()
+                aa=np.array( [[f.readline().split()[2:8] 
                              for n in range(self.num_wann)] 
                                 for m in range(self.num_wann)],dtype=float)
-            self.AA_R[:,:,ir,:]=(aa[:,:,0::2]+1j*aa[:,:,1::2]).transpose( (1,0,2) ) /self.Ndegen[ir]
+                self.AA_R[:,:,ir,:]=(aa[:,:,0::2]+1j*aa[:,:,1::2]).transpose( (1,0,2) ) /self.Ndegen[ir]
+            self.wannier_centers_cart_auto =  np.array( np.diagonal(self.AA_R[:,:,self.iR0,:],axis1=0,axis2=1).transpose() )
```
jaemolihm, line 83: .T instead of transpose. Also do you need np.array?

wannierberri/__system_w90.py
```diff
@@ -65,21 +65,20 @@ def __init__(self,seedname="wannier90",
...
+            ws_map=ws_dist_map_gen(self.iRvec,chk.wannier_centers, chk.mp_grid,self.real_lattice, npar=npar)
         
         eig=EIG(seedname)
         if self.getAA or self.getBB:
             mmn=MMN(seedname,npar=npar)
 
         kpt_mp_grid=[tuple(k) for k in np.array( np.round(chk.kpt_latt*np.array(chk.mp_grid)[None,:]),dtype=int)%chk.mp_grid]
         if (0,0,0) not in kpt_mp_grid:
-            raise ValueError("the grid of k-points read from .chk file is not Gamma-centrered. Please, use Gamma-centered grids in the ab initio calculation")
+            raise ValueError("the grid of k-points read from .chk file is not Gamma-centerred. Please, use Gamma-centered grids in the ab initio calculation")
```
jaemolihm, line 81: rred

wannierberri/__system.py
```diff
@@ -284,23 +289,130 @@ def set_symmetry(self,symmetry_gen=[]):
...
         """ 
         With self.use_wcc_phase=True it is R+tj-ti. With self.use_wcc_phase=False it is R. [m,n,iRvec] (Cartesian)
         """
-        wannier_centres = self.wannier_centres_cart
-        w_centres = np.array([[j-i for j in wannier_centres] for i in wannier_centres])
+        wannier_centers = self.wannier_centers_cart
+        w_centers = np.array([[j-i for j in wannier_centers] for i in wannier_centers])
         if self.use_wcc_phase:
-            return self.iRvec.dot(self.real_lattice)[None,None,:,:]+ w_centres[:,:,None,:]
+            return self.iRvec.dot(self.real_lattice)[None,None,:,:]+ w_centers[:,:,None,:]
```
jaemolihm, line 304: Remove two lines outside if, and use self.cRvec[None,None, :, :] + self.diff_wcc_cart[:, :, None, ].

wannierberri/__system.py
```diff
@@ -284,23 +289,130 @@ def set_symmetry(self,symmetry_gen=[]):
...
-    #def cRvec(self):
-    #    return self.iRvec.dot(self.real_lattice)
+    @lazy_property.LazyProperty
+    def cRvec(self):
+        return self.iRvec.dot(self.real_lattice)
 
     @lazy_property.LazyProperty
     def cRvec_wcc(self):
         """ 
         With self.use_wcc_phase=True it is R+tj-ti. With self.use_wcc_phase=False it is R. [m,n,iRvec] (Cartesian)
```
jaemolihm, line 299: m n or i j. Also “,iRvec,a]”

jaemolihm (commented):
I finished the review. I left some comments.

wannierberri/__Data_K.py
```diff
@@ -47,26 +48,19 @@ def __init__(self,system,dK,grid,Kpoint=None,npar=0,fftlib='fftw',npar_k=1 ):
         self.ksep = system.ksep
         self.use_wcc_phase=system.use_wcc_phase
         if self.use_wcc_phase:
-            self.wannier_centres_reduced=system.wannier_centres_reduced
-            self.wannier_centres_cart=system.wannier_centres_cart
+            # if we consistently "ignore" the factor exp(ik*(tj-ti)) bothin the Wannier Hamiltonian, and in the  
+            # thransformation to Hamiltonian gauge, the result will be correct again.
+            # We only need wcc to evaluate "," derivatives (to be used in cRvec_wcc)
```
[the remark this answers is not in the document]
jaemolihm: This is the essential part of the new implementation. Could you make it explicit here that when calculating the derivatives we multiply R if False and R+ti - tj if True, and this choice is done inside cRvec_wcc?Btw, the name crvec_wcc is somewhat confusing because it is just crvec if use_wcc=false.

wannierberri/__system.py
jaemolihm: If contiguity is a problem, you can use https://numpy.org/doc/stable/reference/generated/numpy.ascontiguousarray.html

wannierberri/__system.py
```diff
@@ -284,23 +289,130 @@ def set_symmetry(self,symmetry_gen=[]):
...
+
+            # not sure if the following is correct (Stepan)
+            if hasattr(self,'SA_R'):
+                assert hasattr(self,'SS_R') , "if you use SA_R, you need also SS_R"
+                SA_R_new  =  self.SA_R.copy() - self.SS_R[:,:,:,:,None]*self.wannier_centers_cart[None,:,None,None,:]
+            if hasattr(self,'SHA_R'):
+                assert hasattr(self,'SH_R') , "if you use SA_R, you need also SH_R"
+                SHA_R_new  =  self.SHA_R.copy() - self.SS_R[:,:,:,:,None]*self.wannier_centers_cart[None,:,None,None,:]
+
+            for X in ['AA','BB','CC','SA','SHA']:
```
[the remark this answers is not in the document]
stepan-tsirkin: Mainly because to modify CC, we need the old values of BB. But to make sure we do not run into such collisions for other quantities, we generally do modificarions using all XX_R's , and in the end assign XX_R=XX_R_new

wannierberri/__system.py
stepan-tsirkin: ok

wannierberri/__system.py
stepan-tsirkin: ok

wannierberri/__system_tb.py
stepan-tsirkin: right

wannierberri/__system.py
stepan-tsirkin: right

tests/test_integrate.py
```diff
@@ -88,12 +95,15 @@ def compare_quant(quant):
...
+    check_integrate(system_Fe_W90 , quantities_Fe , fout_name="berry_Fe_W90" , suffix="" , Efermi=Efermi_Fe , comparer=compare_energyresult,compare_smooth = False ,
+            extra_precision = {"Morb":-1e-6})
 
 
 def test_Fe_wcc(check_integrate,system_Fe_W90_wcc, compare_energyresult,quantities_Fe,Efermi_Fe):
     """Test anomalous Hall conductivity , ohmic conductivity, dos, cumdos"""
-    check_integrate(system_Fe_W90_wcc , quantities_Fe , fout_name="berry_Fe_W90" , suffix="wcc" , Efermi=Efermi_Fe , comparer=compare_energyresult )
+    check_integrate(system_Fe_W90_wcc , quantities_Fe , fout_name="berry_Fe_W90" , suffix="wcc" , Efermi=Efermi_Fe , comparer=compare_energyresult,
+            extra_precision = {"Morb":-5e-2})  # the wcc gives quite a notable error, do not know why yet
+    compare_energyresult( "berry_Fe_W90", "Morb-wcc",  0 , suffix_ref="Morb-wcc" ,precision=-1e-8, compare_smooth = False )
```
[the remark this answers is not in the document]
stepan-tsirkin: done

tests/test_integrate.py
```diff
@@ -67,7 +69,12 @@ def Efermi_Haldane():
 
 @pytest.fixture(scope="session")
 def quantities_Fe():
-    return  ['ahc','ahc_ocean','dos','cumdos'  ,'conductivity_ohmic','conductivity_ohmic_fsurf']
+    return  ['ahc','ahc_ocean','dos','cumdos'  ,'conductivity_ohmic','conductivity_ohmic_fsurf','Morb']
+
+@pytest.fixture(scope="session")
+def quantities_Fe_frozen():
+    return  ['Morb',]
```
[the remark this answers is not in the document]
stepan-tsirkin: yes, I meant it for smth, but did not use so far. removing for now.

tests/test_integrate.py
```diff
@@ -22,6 +22,8 @@ def _inner(system,quantities,fout_name,Efermi,comparer,
                grid_param={'NK':[6,6,6],'NKFFT':[3,3,3]},additional_parameters={},adpt_num_iter=0,
                suffix="", suffix_ref="",
                extra_precision={},
+               precision = -1e-8 ,
+               compare_smooth = False,
```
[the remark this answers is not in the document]
stepan-tsirkin: My fault, it was a temporary thing, that I forgot to clean up. Now `compare_smooth = True` everywhere. Tests pass.

stepan-tsirkin:
Thank you @jaemolihm . Yes, I did not change reference data, nor loosen precision. I am merging now.

jaemolihm:
It seems that some of the comments are not dealt with. (They are "hidden", you can click "load more" to see them.) I will make a PR to implement my comments.

PR #82, opened 2021-10-13
Fix bug in Path with k_nodes argument

jaemolihm:
I fixed a bug in Path initialized with `k_nodes` argument.

- [x] Add test

reproducer:
```python3
import os
import wannierberri as wb
import pythtb as ptb
import numpy as np

def HaldanePTB(delta,t1,hop2,phi):
    lat=[[1.0,0.0],[0.5,np.sqrt(3.0)/2.0]]
    orb=[[1./3.,1./3.],[2./3.,2./3.]]
    haldane=ptb.tb_model(2,2,lat,orb)
    haldane.set_onsite([-delta,delta])
    haldane.set_hop(t1, 0, 1, [ 0, 0])

    return haldane


haldane = HaldanePTB(2,1,1/3,np.pi/10)
syst=wb.System_PythTB(haldane,berry=True,morb=True)

k_nodes = [[0.0, 0.0, 0.0], [0.5, 0.5, 0.5]]
pp = wb.Path(syst, k_nodes=k_nodes, nk=3)
print(pp.labels)
print(pp.K_list)
```


stepan-tsirkin:
@jaemolihm , will you be able to add a small test soon? You may just add to your reproducing example 

```
assert pp.labels == {0: '1', 2: '2'}   ,  "some comment if it fails"
assert np.all( pp.K_list == np.array([[0. ,  0.  , 0.  ], [0.25, 0.25 ,0.25], [0.5 , 0.5 , 0.5 ]])) ,  "some comment if it fails"
```

I was planning to make a release 0.10.1 , which already includes several bugfixes and the FPLO interface (#91) .

jaemolihm:
Sorry for the delay. I found several other issues in `Path`, so I was trying to fix them too. But I think you can just merge this. I opened an issue to track other problems #99.

jaemolihm:
Also, it could be v0.11.0, as we have many new features. (Usually, the version number means `MAJOR.MINOR.PATCH`, where PATCH is for bugfix. https://semver.org/) But this is not a strict rule, and we are at v0, so you can do anything you want.

tests/test_path.py
```diff
@@ -0,0 +1,24 @@
...
+        haldane=ptb.tb_model(2,2,lat,orb)
+        haldane.set_onsite([-delta,delta])
+        haldane.set_hop(t1, 0, 1, [ 0, 0])
+
+        return haldane
+
+
+    haldane = HaldanePTB(2,1,1/3,np.pi/10)
+    syst=wb.System_PythTB(haldane,berry=True,morb=True)
+
```
stepan-tsirkin, line 18: We already have a fixture 

 
 ​system_Haldane_PythTB​

Is there a reason not to use it here?

stepan-tsirkin (changes requested):

tests/test_path.py
jaemolihm: No, I didn't know it. I'll fix it soon.

stepan-tsirkin (approved):

PR #85, opened 2021-11-08
Static Spin Hall conductivity in FermiOcean

jaemolihm:
This PR implements static spin Hall conductivity calculation via FermiOcean. The quantities are called `"shc_qiao"` and `"shc_ryoo"`. (The Kubo versions are called "opt_SHCqiao", "opt_SHCryoo")

I added a test that compares kubo and FermiOcean implementations for the static case.


jaemolihm:
I have some questions on the new FermiOcean implementation.

1. Should the definition of the spin current matrix go to `covariant_formulak.py` or `data_K.py`?
2. Are objects re-calculated every time? For example, `frml.Omega` is used in various quantities. Is it recalculated, or is it stored somewhere?
3. Should `parity_I` and `parity_TR` defined for all real-space operators? (For spin Hall: SH_R, SA_R, SHA_R, SHR_R) What does `return None` mean there?

stepan-tsirkin:
@functools.lru_cache



> I have some questions on the new FermiOcean implementation.
> 
>     1. Should the definition of the spin current matrix go to `covariant_formulak.py` or `data_K.py`?

Generally, I would like not to overload `data_K`, so the definition of formulae should go to `covariant_formulak.py`, or in a separate file if you want.

> 
>     2. Are objects re-calculated every time? For example, `frml.Omega` is used in various quantities. Is it recalculated, or is it stored somewhere? In general, what should be done, if any, to avoid calculating the same thing multiple times?

Yes, so far they are recalculated. Assuming that the heavy operations are to get the ingredients, which are stored in the `data_K._covariant_quantities` , this should not be a problem, but not sure generally. I did not profile the code properly yto see what takes time or not.

But this is only about __init__ methods. On the other hand the outputs of  `.nn` and 'ln` methods are never stored.  We can store them with a `@functools.lru_cache` decorator, but that may take some big memory. Maybe, saving the `.nn` but not `.ln` method could be a good compromise.  We need to estimate the pros and cons. Or, just make a maxsize an input parameter (`None` will mean infinite memoization) and leave this responsibility to the user. 

Anyway, if needed, that should be done as a separate PR

> 
>     3. Should `parity_I` and `parity_TR` defined for all real-space operators? (For spin Hall: SH_R, SA_R, SHA_R, SHR_R) What does `return None` mean there?

These values are used only if the corresponding Matrix_ln formula is used as itself with a trace() method, or as part of a `FormulaProduct`, if the `TRodd` and `Iodd` methods are not re-defined.  So, for `'D', 'AA', 'BB', 'CCab'` the parity_I and parity_TR are never used. 

So, if a formula may be used directly, e.g. in `tabulate`, then the parities should be defined (to know the transformations under k->-k.

jaemolihm:
Thanks. Then, I will move the spin current definition to `covariant_formulak.py` soon and ask for review again.

jaemolihm:
@stepan-tsirkin I moved spin current definition to `covariant_formulak`. The code is ready for review.

One question: Is it okay to use `data_K.Xbar` in `covariant_formulak`?

stepan-tsirkin (approved):
Yes, that is fine. @jaemolihm  you may merge. And then, please, reflect the changes in the documentation.

jaemolihm:
I renamed `shc_*` to `shc_static_*`.

PR #86, opened 2021-11-10
fixes in models, tabulating morb and der_morb

stepan-tsirkin:
This PR implements the folowing:

* tabulating of `morb` and `Der_morb` (incl tests)
* in `models.py` the impott of `pythtb` and `tbmodels` are moved inside the corresponding functions. So, the module may be imported even if `pythtb` and ``tbmodels` are not installed
* small fixes and comments in the models.


wannierberri/models.py
```diff
@@ -1,14 +1,31 @@
 """Here we define some models, that can be used to test the code, or just to play around"""
 
-import tbmodels
-import pythtb 
 import numpy as np
 
 def Haldane_tbm(
     delta=0.2,
     t=-1.0,
```
jaemolihm, line 7: Hop1 (or hop) and hop2 seems better than t and hop2. Same for pther models.

jaemolihm (changes requested):

PR #91, opened 2021-11-14
interface to FPLO code

stepan-tsirkin:
System_fplo allows to use Wannier functions from the [FPLO](https://www.fplo.de/) code.  

Only the internal terms may be evaluated.

A test is included for bcc Fe (from a 2x2x2 grid)

An associated [PR to documentation](https://github.com/wannier-berri/wannier-berri-org/pull/7)


wannierberri/__system_fplo.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Notes
+    -----
+    see also  parameters of the :class:`~wannierberri.System` 
+    """
+
+    def __init__(self,hamdata="+hamdata",**parameters):
+
+        self.set_parameters(**parameters)
+        if self.morb : raise ValueError("System_tb class cannot be used for evaluation of external terms in orbital magnetic moments")
```
jaemolihm, line 43: System tb -> fplo

wannierberri/__system_fplo.py
```diff
@@ -0,0 +1,118 @@
...
+        if self.berry : raise ValueError("System_tb class cannot be used for evaluation of external terms of berry curavture")
```
jaemolihm, line 44: Tb -> fplo

wannierberri/__system_fplo.py
jaemolihm, line 43: Stystem tb -> fplo

wannierberri/__system_fplo.py
```diff
@@ -0,0 +1,118 @@
...
+#        if self.spin : raise ValueError("System_tb class cannot be used for evaluation of spin properties")
```
jaemolihm, line 45: If spin can be used, you can delete this comment

jaemolihm (changes requested):
Looks good to me.

Do You use wcc true in the test because FPLO does not provide the AA matrix? If so, could you mention thisin system fplo ? (Maybe also a warning (not an error) if use wcc phase is false?).

Or is the AA matrix automatically set to be diagonal using the wann centers when use wcc phase is false? If this is the case, there is no need to do what I said. Then you could also test the non wcc system with the same reference data.

stepan-tsirkin:
Thanks for comments, @jaemolihm 
Yes, I allowed for external terms with diagonal AA_R matrix (like for TB models) and print a warning that it is not recommended to do so. 

Additional tests pass, and also note that FPLO gives perfectly symmetric WFs. A test checks this also.

Any more comments?

jaemolihm:
Looks good.

jaemolihm (approved):

jaemolihm:
@stepan-tsirkin , I saw a paper on arxiv about Wannier functions in FPLO code : "Symmetry Conserving Maximally Projected Wannier Functions" https://arxiv.org/abs/2111.09652.

I have no experience with FPLO code, so I have some questions. Do you know the following?
1. Is the FPLO WFs just the projection-only WFs of Wannier90 (i.e. `num_iter=0`, `dis_num_iter=0`?)
2. Is there anything one should learn from FPLO in terms of Wannier function method development? Or is it just a new implementation of the same algorithms?

stepan-tsirkin:
@jaemolihm  , i actually have no experience with the FPLO. Neithew I have seen its source code.

Just some collaborators asked me for the interface and it was easy to do such. They provided a sample of input for tests, so I did not even run FPLO myself. 

But from the test it really comes out that the WFs are symmetric. I also sow the paper, but did not read it in detail.

PR #94, opened 2021-11-19
degen_thresh and degen_Kramers

stepan-tsirkin:
Now `degen_thresh` is an input parameter for all FermiOcean quantities. 
Also add a parameter `degen_Kramers` which makes the bands (0,1) , (2,3), ...considered always as degenerate, regardless of actual splitting. This may be used only for PT-symmetric systems

Included a test , whith a model  that slightl;y breaks PT symmetyry, and check that `degen_Kramers=True` and `degen_thresh=0.05` give the same results.


wannierberri/models.py
```diff
@@ -152,3 +152,107 @@ def Chiral(
...
+
+    Parameters
+    -----------
+    nx : float
+        x-component of the Néel vector
+    ny : float
+        y-component of the Néel vector
+    nz : float
+        z-component of the Néel vector
+    t : float
```
jaemolihm, line 181: Use hop1 and hop2 to be consistent with other models

tests/test_integrate.py
```diff
@@ -354,3 +371,41 @@ def test_Chiral_tetra(check_integrate,system_Chiral,compare_energyresult,quantit
...
+               grid_param={'NK':[10,10,1], 'NKFFT':[5,5,1]} )
+
+    for quant in quantities_CuMnAs_2d:
+        for tetra in True,False:
+            degen = degen_param[0]
+            qfull1 = f"{quant}^tetra={tetra}_{degen[0]}={degen[1]}"
+            degen = degen_param[1]
+            qfull2 = f"{quant}^tetra={tetra}_{degen[0]}={degen[1]}"
+            data1=result.results.get(qfull1).data
+            data2=result.results.get(qfull1).data
```
jaemolihm, line 402: qfull2?

tests/test_integrate.py
```diff
@@ -354,3 +371,41 @@ def test_Chiral_tetra(check_integrate,system_Chiral,compare_energyresult,quantit
...
+            assert data1.shape == data2.shape
+            assert np.all( np.array(data1.shape[1:]) == 3)
+            assert np.all( np.array(data2.shape[1:]) == 3)
+            precision = 1e-14*np.max(abs(data1))
+            print (qfull1,qfull2)
```
jaemolihm, line 407: Delete this line

jaemolihm (changes requested):

jaemolihm:
I left some minor comments. Looks hood otherise.

jaemolihm:
Instead of writing degen thresh = 1e-4, degen kramers false in every quantities, what about passing **parameters to FermiOcean?

stepan-tsirkin:
> Instead of writing degen thresh = 1e-4, degen kramers false in every quantities, what about passing **parameters to FermiOcean? 

The problem is that `**parameters` are passed to the formula, 
Well, if `**parameters` includes `degen_thresh` , and error will **not** occur in a Formula, just a useless property will appear in it. Athough it is not nice
But if we pass `external_terms` to `FermiOcean` trough `**paramerters` there will be an error. 

Do you know an elegant solution to this problem? If you have, you can mkake another  PR in future.

>Also **kwargs (keyword arguments) is more idiomatic in python.

For the reason above, maybe rename it to `**formula_kwargs` , to distinguish from other parameters.

jaemolihm:
I see. I can't think of a good solution now; I'll make a PR if I find one.

PR #104, opened 2021-11-26
fixes for tabulation along path

stepan-tsirkin:
Work  in progress

TODO

- [x] add a test
- [x] use repr/str method for a grid/pasth


wannierberri/__evaluate.py
```diff
@@ -124,7 +124,11 @@ def evaluate_K(func,system,grid,fftlib='fftw',
             file_Klist+=".pickle"
     cnt_exclude=0
 
-    print ("using NKdiv={}, NKFFT={}, NKtot={}".format( grid.div,grid.FFT,grid.dense))
+    # TODO : implement and use the repr() or str() method of grid/path
+    try:
+        print ("using NKdiv={}, NKFFT={}, NKtot={}".format( grid.div,grid.FFT,grid.dense))
```
Liu-Xiaoxiong, line 129: Sorry, I found this small bug before, and fixed in PR sym_wann. Then forget it totally

Like give the div of the path in path.py
```
        self.div = np.shape(self.K_list)[0]
```

Liu-Xiaoxiong (commented):
I suggest print the NKdiv as the number of points in the path Klist. What do you think?

jaemolihm:
https://github.com/wannier-berri/wannier-berri/blob/a0d8e41e2a215e7f5c845f342d319f876961461f/wannierberri/__tabulate.py#L324-L326

Should `quantity='berry'` be `quantity=quantity`?

jaemolihm:
Also, these two lines should be `for ib in range(E.shape[1]):`, because `iband` is already used in `get_data`.
https://github.com/wannier-berri/wannier-berri/blob/a0d8e41e2a215e7f5c845f342d319f876961461f/wannierberri/__tabulate.py#L310-L311
https://github.com/wannier-berri/wannier-berri/blob/a0d8e41e2a215e7f5c845f342d319f876961461f/wannierberri/__tabulate.py#L328-L329

wannierberri/__evaluate.py
AbhirupRK: Where do I put this line?

wannierberri/__tabulate.py
```diff
@@ -307,7 +307,7 @@ def plot_path_fat(self,
             Emax=E.max()+0.5
 
         klineall=[]
-        for ib in iband:
+        for ib in range(E.shape[1]):
```
jaemolihm, line 310:
```suggestion
        for ib in range(len(iband)):
```

(Sorry, I know I suggested `range(E.shape[1])` but `range(len(iband))` seems more informative)

wannierberri/__tabulate.py
```diff
@@ -322,10 +322,10 @@ def plot_path_fat(self,
             kmax=kline.max()
 
         if quantity is not None:
-            data=self.get_data(quantity='berry',iband=iband,component=component)
+            data=self.get_data(quantity=quantity,iband=iband,component=component)
             print ("shape of data",data.shape)
             if mode=="fatband" :
-                for ib in iband:
+                for ib in range(E.shape[1]):
```
jaemolihm, line 328:
```suggestion
                for ib in range(len(iband)):
```

tests/test_tabulate.py
```diff
@@ -19,17 +19,17 @@
...
-            return ["-"+a for a in "z"]
+            return [a for a in "z"]
         if quantity in ["Der_berry","Der_morb"]:
-            return  ["-"+a+b for a in "xyz" for b in "xyz"]
+            return  [a+b for a in "xyz" for b in "xyz"]
         if quantity == "omega2" :
-    	    return ["-zz"]
+    	    return ["zz"]
         raise ValueError(f"unknown quantity {quantity}")
     return _inner
```
jaemolihm, line 34: This can be a simple function, not a fixture. (Then, we do not need `_inner`)

A fixture-function is needed if one needs to test something inside it, such as in the comparers.

tests/test_path.py
```diff
@@ -37,3 +41,45 @@ def test_path(system_Haldane_PythTB):
                                                   [3, 3, 3],
                                                   [4, 4, 4],
                                                   [5, 5, 5],]) / 10), "path.K_list is wrong"
+
+
+    quantities = quantities_tab
```
jaemolihm, line 46: Separate this part to a new function `test_path_tabulate` (because `test_path` focuses on the construction of the path itself).

tests/test_path.py
```diff
@@ -37,3 +41,45 @@ def test_path(system_Haldane_PythTB):
...
+    tab_result_ref = pickle.load(open(os.path.join(REF_DIR, filename),"rb") )
+
+    for quant in ["E"]+quantities:
+        for comp in get_component_list(quant):
+            data     =     tab_result.results.get(quant).get_component(comp)
+            data_ref = tab_result_ref.results.get(quant).get_component(comp)
+            assert data == approx(data_ref), (f"tabulation along path gave a wrong result for quantity {quant} component {comp} "+
+                "with a maximal difference {}".format(max(abs(data-data_ref)))   )
+
+    ## only checks that the plot runs without errors, not checkoing the result of the plot
```
jaemolihm, line 74: checking

tests/test_path.py
```diff
@@ -37,3 +41,45 @@ def test_path(system_Haldane_PythTB):
...
+                ibands = [0],
+                use_irred_kpt = True , symmetrize = True, # should have no effect, but will check the cases and give a warning
+#                parameters_K = parameters_K,
+#                frmsf_name = None,
+#                degen_thresh = degen_thresh, degen_Kramers = degen_Kramers
+                )
+
+    filename = "path_tab.pickle"
+    fout = open(os.path.join(OUTPUT_DIR, filename),"wb")
+    pickle.dump(tab_result,fout)
```
jaemolihm, line 64: Do we need to dump `tab_result`?

tests/test_path.py
```diff
@@ -37,3 +41,45 @@ def test_path(system_Haldane_PythTB):
...
+    tab_result_ref = pickle.load(open(os.path.join(REF_DIR, filename),"rb") )
```
jaemolihm, line 65: Instead of saving the whole results object as pickled file (which can break when we change the structure of result class), I would prefer to save only the data.

For example:
```python3
data = {}
for quant in ["E"]+quantities:
    for comp in get_component_list(quant):
        data[f"{quant}.{comp}"] = tab_result.results.get(quant).get_component(comp)

with open(os.path.join(REF_DIR, filename), "rb") as f:
    pickle.dump(data, f)
```

jaemolihm (changes requested):

tests/test_tabulate.py
jaemolihm: Also, why do you compare only z component for Berry, V  morb?

tests/test_path.py
stepan-tsirkin: Done. Made construction as a ixture.

tests/test_path.py
stepan-tsirkin: Done

tests/test_tabulate.py
stepan-tsirkin: Changed to function, changed "z" yo "xyz"

tests/test_path.py
stepan-tsirkin: Just to have smth to put in te reference folder. But it does not hurt during test

Issue #99, opened 2021-11-22
Issues in Path class


jaemolihm:
One is fixed in #82.

There are other issues, which I will describe here later.

stepan-tsirkin:
Is the other one related to #104 ? Or is there another problem?

jaemolihm:
Sorry for the delay. No, it's another problem.

The problem is that the intermediate nodes are duplicated (` [0.5   0.5   0.5  ]` occurs twice). I'm not sure this is a bug or a feature, though.

Code:
```python3
import os
import wannierberri as wb
import pythtb as ptb
import numpy as np

def HaldanePTB(delta,t1,hop2,phi):
    lat=[[1.0,0.0],[0.5,np.sqrt(3.0)/2.0]]
    orb=[[1./3.,1./3.],[2./3.,2./3.]]
    haldane=ptb.tb_model(2,2,lat,orb)
    haldane.set_onsite([-delta,delta])
    haldane.set_hop(t1, 0, 1, [ 0, 0])

    return haldane


haldane = HaldanePTB(2,1,1/3,np.pi/10)
syst=wb.System_PythTB(haldane,berry=True,morb=True)

k_nodes = [[0.0, 0.0, 0.0], [0.5, 0.5, 0.5], [0.0, 0.0, 0.5]]
pp = wb.Path(syst, k_nodes=k_nodes, nk=5)
print(pp.labels)
print(pp.K_list)
```

Output: 
```
[jmlim@im0 wannier-berri]$ python3 a.py 
R=0 found at position(s) [[0]]
[]
NOT using ws_dist
Number of wannier functions: 2
Number of R points: 1
Recommended size of FFT grid [1 1 1]
Reading the system from PythTB finished successfully
{0: '1', 4: '2', 5: '2', 9: '3'}
[[0.    0.    0.   ]
 [0.125 0.125 0.125]
 [0.25  0.25  0.25 ]
 [0.375 0.375 0.375]
 [0.5   0.5   0.5  ]
 [0.5   0.5   0.5  ]
 [0.375 0.375 0.5  ]
 [0.25  0.25  0.5  ]
 [0.125 0.125 0.5  ]
 [0.    0.    0.5  ]]
```

stepan-tsirkin:
Well, it does not hurt much, so it is not a bug.  but it does not help much either, so it is not a feature. 

If you have time, you may correct it. To exclude repeating points.

jaemolihm:
[...]

PR #110, opened 2021-12-07
Remove node duplication in Path

jaemolihm:
Fix the k-point duplication issue mentioned in https://github.com/wannier-berri/wannier-berri/issues/99#issuecomment-987771817

Added tests


stepan-tsirkin (approved):
All fine, thanks. Please, merge.

PR #111, opened 2021-12-08
Add option to disable printing of K points

jaemolihm:
Added `print_Kpoints` argument to `integrate` and `tabulate`. When set to `False`, the list of K points are not printed. The default is `True`.

I added this option because when I have many K points, the output becomes very long.

### DIFF
--- a/wannierberri/__main.py
+++ b/wannierberri/__main.py
@@ -114,6 +114,7 @@ def integrate(system,grid,Efermi=None,omega=None, Ef0=0,
                         use_irred_kpt = True, symmetrize = True,
                         fout_name="wberri",restart=False,fftlib='fftw',suffix="",file_Klist="Klist",
                         parallel = None,
+                        print_Kpoints = True,
                         parameters={},parameters_K={},specific_parameters={} ):
     """
     Integrate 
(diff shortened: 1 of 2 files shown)


wannierberri/__main.py
```diff
@@ -149,6 +150,8 @@ def integrate(system,grid,Efermi=None,omega=None, Ef0=0,
         evaluate only symmetry-irreducible K-points
     symmetrize : bool
         symmetrize the result (always `True` if `use_irred_kpt == True`)
+    print_Kpoints : bool
+        print the list of K points (default: True)
```
stepan-tsirkin, line 154: There is generally no need to write the defaults in the docstrings, because when sphinx generates documentation it takes defaults from the method definition. Thus it creates a risk to make the docstring to become in consistant with the actual definition of the defaults.

wannierberri/__main.py
```diff
@@ -241,6 +246,8 @@ def tabulate(system,grid, quantities=[], user_quantities = {},
         if not None, the results are also printed to text files, ready to plot by for `FermiSurfer <https://fermisurfer.osdn.jp/>`_
     parallel : :class:`~wannierberri.Parallel`
         object describing parallelization scheme
+    print_Kpoints : bool
+        print the list of K points (default: True)
```
stepan-tsirkin, line 250: same here: no need to print defaults.

stepan-tsirkin (changes requested):

jaemolihm:
Thanks, I fixed them.

stepan-tsirkin (approved):
Great!

PR #113, opened 2021-12-10
Fix ray cluster

stepan-tsirkin:
There was a bug, reported by email to me by  Yibo Wang

looks like nobody used the "cluster" option of ray recently,

akso updated the dostring of `Parallel()` and  the tutorial


Liu-Xiaoxiong:
There is another thing about ray is: 
Ray will store tmp files to the default address. If we run it on the cluster the default might be somewhere that only administrators have permission to delete. 
The tmp files are very large if you run some heavier calculations.
(I have filled up that folder recently!)

How about setting the address where store tmp file under the folder of the run script.

What do you think?

stepan-tsirkin:
@Liu-Xiaoxiong , I did not know it.  If you know how to do it, you can add a commit to this PR. Is it done via parameter of ray.init() ?

Liu-Xiaoxiong:
> @Liu-Xiaoxiong , I did not know it. If you know how to do it, you can add a commit to this PR. Is it done via parameter of ray.init() ?

Can be done by parameters of ray.init(),  in the document. I will figure it out and add a commit later.

stepan-tsirkin:
> Can be done by parameters of ray.init(), in the document. I will figure it out and add a commit later.

Then it may be added similar to https://github.com/wannier-berri/wannier-berri/blob/3c813d018c973b172d9549438791a997b91c0518/wannierberri/__parallel.py#L52-L60

So that by default it is in the current folder, and there is an option to overwrite.

But what is in those tmp files? Wouldn't placing them in current folder affect performance? Could you provide a link to docs?

Liu-Xiaoxiong:
@stepan-tsirkin 
The link to the doc about temporary files is : https://docs.ray.io/en/stable/tempfile.html
See the NOTE at the bottom. They also think /tmp is not a good choice. 

I don't know if it affects the performance or not.
use 
'''
ray_init={'_temp_dir': os.getcwd()} 
'''
Can change it. But there is another issue from my test.

wannierberri/__parallel.py
```diff
@@ -48,6 +47,18 @@ def __init__(self,
...
+
+                def set_opt(opt,def_val):
+                    if opt not in ray_init:
+                        ray_init_loc[opt] = def_val()
+                    else:
+                        print (f"WARNING: the ray cluster will use '{ray_init[opt]}' provided in ray_init")
+
+                set_opt('address'          , lambda : 'auto')
+                set_opt('_node_ip_address' , lambda : os.environ["ip_head"].split(":")[0])
+                set_opt('_redis_password'  , lambda : os.environ["redis_password"])
```
jaemolihm, line 60: Why did you use lambda function instead of just passing a variable?

tests/conftest.py
```diff
@@ -39,14 +38,33 @@ def parallel_ray():
     # of return, and add parallel.shutdown() after the yield statement.
     # See https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended
     # Currently, only a single ray setup is used, so this is not a problem.
+    "first we just check that the initialization works with some dummy ray_init parameters"
```
jaemolihm, line 41: If this is a comment, it would be better to use `#`

jaemolihm (approved):
I have little experience with ray, so I cannot say much. I only left some minor comments.

stepan-tsirkin:
@Liu-Xiaoxiong could you transform your post into a comment on the webpage in the parallel section, incl links to ray doc and the bugreport

It says that the tmp files are for logging and sharing object store with **other** programs. So, I think the performance of the python code standalone should not be affected. What I was worrying is that different directories may be physically located on different drives with different connection and read/write speed. But looks like it does not really matter.

tests/conftest.py
stepan-tsirkin: ok

wannierberri/__parallel.py
stepan-tsirkin: added a comment in the code:

```
                # The follwoing is done for testing, when __init__ is called with `cluster = True`,
                # but no actual ray cluster was initialized (and hence the needed environmental variables are not set
```

Issue #114, opened 2021-12-11
Failed to install `wannierberri` using python version >= 3.9 on Ubuntu 20.04.3 LTS.


hongyi-zhao:
On Ubuntu 20.04.3 LTS, I try to install `wannierberri` using python 3.9.1 and 3.10.1 managed by [`pyenv`](https://github.com/pyenv/pyenv), but both failed:

```
# Python 3.10.1 failed with a similar message.
$ pyenv shell 3.9.1
$ pyenv virtualenv --system-site-packages wannierberri
$ pyenv shell wannierberri
$ pip install -U pip
$ pip install wannierberri
#Or use the following command to install from a locally cloned git repository
$ pip install -e .
```
However, the installation step described above failed with the following error:
```
$ pip3 install wannierberri
Looking in indexes: https://pypi.tuna.tsinghua.edu.cn/simple
Collecting wannierberri
  Downloading https://pypi.tuna.tsinghua.edu.cn/packages/48/92/ae5d18e1db86ebe9a0b32a905d2f1ef67f9d972a2e87266f0aa42bd8541a/wannierberri-0.11.0-py3-none-any.whl (117 kB)
     |████████████████████████████████| 117 kB 4.2 MB/s            
Requirement already satisfied: matplotlib in /home/werner/.pyenv/versions/3.9.1/envs/datasci/lib/python3.9/site-packages (from wannierberri) (3.4.2)
Requirement already satisfied: scipy>=1.0 in /home/werner/.pyenv/versions/3.9.1/envs/datasci/lib/python3.9/site-packages (from wannierberri) (1.6.3)
Requirement already satisfied: packaging>=20.8 in /home/werner/.pyenv/versions/3.9.1/envs/datasci/lib/python3.9/site-packages (from wannierberri) (20.9)
Requirement already satisfied: numpy in /home/werner/.pyenv/versions/3.9.1/envs/datasci/lib/python3.9/site-packages (from wannierberri) (1.20.3)
Requirement already satisfied: pyfiglet in /home/werner/.pyenv/versions/3.9.1/envs/datasci/lib/python3.9/site-packages (from wannierberri) (0.8.post1)
Requirement already satisfied: colorama in /home/werner/.pyenv/versions/3.9.1/envs/datasci/lib/python3.9/site-packages (from wannierberri) (0.4.1)
Collecting pyFFTW>=0.12.0
[...]

PR #124, opened 2022-01-06
Add python3.9 to CI

jaemolihm:
As the pyFFTW issue (#114) is fixed (on the pyFFTW side), python3.9 can be safely added to CI.

Python3.6 reached its end of life (https://endoflife.date/python), so we could remove it from CI, but I just left it because it does little (or no) harm.


stepan-tsirkin (approved):
Good

Issue #125, opened 2022-01-07
Order in passing parameters.


stepan-tsirkin:
The number of parameters that we use in calculation is growing, and we are now using a complicated system to determine which parameters shopuld be passed where. I feel that it is becoming a mess now, both in terms of further implementation and in terms of documentation.

We now have `parameters_K`, `parameters` and `specific_parameters`, and also `quantities` and `user_quantities` . 

I suggest a different scheme. Namely, we will have no pre-defined `quantities`. Instead all calculators will be passed as a sictionary `{'key':calculator}` (similar to `user_quantities`), where `calculator` accepts a single  parameter `data_K`, and nothing more. 

Our pre-defined calculators will be functions of more parameters, so the user before running a calculation has to reduce their number using `functools.partial`. All parameters should be documented in the docstring of the calculator (common parameters may be documented in the docstring of the corresponding module, to avoid multiple repeating of the same). 

In my view, this will make it clearer (both for the user and developer), which parameters go where, and allow 

Explicitly to `integrate` or `tabulate` we should pass only parameters that have to be common for all quantities , such as `grid`, refinement options, `parameters_K` .  even `Efermi` and `omega` should not be passed explicitly to `integrate`

What do you think of this? Do you see any drawbacks of such approach?

jaemolihm:
I strongly agree that the input parameters needs some cleanup. I think what you propose is a good solution. Everything becomes a `Calculator`, so the problem are divided into two smaller pieces (setting up a `Calculator` with proper arguments, and then `run`-ning a calculation).

One possible drawback that come to my mind is that the input argument can be somewhat complicated. Of course, the current situation is also complicated, so changing it is needed. But the following seems a bit too verbose, compared to the current interface `integrate(quantities=["ahc"], ...)` and `tabulate(quantities=["Berry"], ibands=np.arange(4, 10), ...)`. Maybe we can provide a utility that looks like the previous `__integrate` and `__tabulate` interface with minimal, widely (but not always) used arguments (e.g. `tetra`) and returns `calculators` dictionary, so that users who just want simple, well-known quantities (e.g. AHC, Berry dipole, spin Hall, shift current, energy) can just use them? (Something like `get_calculators_with_old_style_syntax(integrate_quantities=["ahc"], tetra=True)` (name should be changed...))
```python
                "ahc_calc":wberri.calculators.AHC(Efermi,tetra=True),
                 "tabulate":wberri.calculators.TabulatorAll({
[...]

PR #127, opened 2022-01-09
`run()`  instead of `integrate()` and `tabulate()`

stepan-tsirkin:
The idea that was raised in #125, developed further.

the new function `run()` substitutes the old `integrate()` and `tabulate()`.  So, it should be possible to integrate and tabulate in one shot.  The idea that every quantity is a `calculator` that is an object that can be called with a single parameter `data_K` and returns a `Result` of some type. 
the `Result` knows by itself how to save it to file, so the `run`

TODO:

- [ ] rewrite all functionalities as `calculators`
- [x] make it work for tabulation
- [ ] design a format for saveing tabukation result as `npz`
- [ ] make sure things work when different Efermi is used for different quantities (or add a check that disables that)

Open Questions:

1. What to do with tabulation under iterations 

    a )  stop after zeros iteration
    b )  save at every iteration npz files containing added points
    c )  give an option to switch between two modes above

2. what to do with smoothing?
    a) add it to definition of calculators?
    b) other

Basically, after implementing tis, the following files will become useless:

- `__main.py`
- `__integrate.py`
- `__tabulate.py`
- `__evaluate.py`

However, we will keep it for some more time for the fans of the old interface. But maybe will remove in the version 1.0.0


stepan-tsirkin:
@jaemolihm , do you have any global suggestions on how to develop this PR ? except the usual issues  - to document and to test everything properly ?

jaemolihm:
I skimmed the code, and I think the implementation is good.

Only a single comment: if I am correct, there is a hierarchy of imports like `fermiocean_dynamic -> calculator -> fermiocean`, while I expect `calculator -> {fermiocean_dynamic, fermiocean}`, so that users (who are not developers) only need to deal with `calculator`.

jaemolihm:
I agree with your plan.

Issue #96, opened 2021-11-20
Code formatting, static code analysis


jaemolihm:
I suggest to use python auto formatter `yapf` to unify the code style, and include `flake8` which checks the code style and programming errors to the CI.

I tried yapf (https://github.com/google/yapf) to automatically format `__kubo.py` and `fermiocean.py` (you can do this by `pip install yapf; yapf -i fermiocean.py`). See https://github.com/wannier-berri/wannier-berri/compare/master...jaemolihm:yapf for changes. It uses the PEP8 style guide https://www.python.org/dev/peps/pep-0008/ by default.
If you think they are good, we can 1) run yapf to the current codes, and 2) add yapf to the git pre-commit hook so that it formats the changes before each commit.
If there is some style that you don't like, maybe it's possible to disable that in the formatter.

`Flake8` can be used to check whether the code is following the style guide. Also, by including `flake8` to the CI, we can find some trivial errors in the PR, such as variables that are not used or undefined. For example, `flake8` gives following comments about `__kubo.py` and `fermiocean.py` (after running yapf):
```linux
jmlihm@DESKTOP-KGFE372:~/program/wannier-berri/wannierberri$ flake8 __kubo.py
__kubo.py:136:121: E501 line too long (147 > 120 characters)
__kubo.py:139:53: W291 trailing whitespace
__kubo.py:270:53: E502 the backslash is redundant between brackets
__kubo.py:271:25: E124 closing bracket does not match visual indentation
__kubo.py:291:13: F841 local variable 'fs' is assigned to but never used

jmlihm@DESKTOP-KGFE372:~/program/wannier-berri/wannierberri$ flake8 fermiocean.py
fermiocean.py:25:121: E501 line too long (141 > 120 characters)
fermiocean.py:26:121: E501 line too long (158 > 120 characters)
fermiocean.py:67:121: E501 line too long (133 > 120 characters)
fermiocean.py:257:51: W291 trailing whitespace
fermiocean.py:288:39: W291 trailing whitespace
fermiocean.py:289:57: W291 trailing whitespace
```
(I took the `.flake8` file from pyscf https://github.com/pyscf/pyscf/blob/master/.flake8. It tells flake8 to ignore or modify some style conventions. For example, the max line length can be increased.)

(The trailing whitespace errors are there because yapf does not format docstrings.)

stepan-tsirkin:
Yes, that will be a good thing, it makes the code more readable 

Maybe I did not like only the following two cases: 

[...]

PR #134, opened 2022-01-30
Setup flake8 static analysis in CI, format some files, fix trivial bugs

jaemolihm:
As the first part of #96, I included static code analysis using flake8 to CI. It prints bugs (e.g. variable not defined but used, typo like `raaise`) as well as formatting problems. Only files in `wannierberri/` are checked. In the future, I will also make files in `tests/` be checked.

I also fixed bugs and formatted some files. I did not touch the files modified in existing PRs, so that this change can be easily merged in those PRs.

Most of the changes are formatting changes (trailing whitespaces, double spaces, ...).

**Nontrivial bug fixes are gathered in this commit**: https://github.com/wannier-berri/wannier-berri/pull/134/commits/0a96889ec192604adf3bc055f9103a9eeb237028. Please check they are correct.


.flake8
```diff
@@ -0,0 +1,15 @@
+[flake8]
+# https://flake8.pycqa.org/en/2.5.5/warnings.html#error-codes
```
stepan-tsirkin, line 2: A more update link is : 
https://flake8.pycqa.org/en/4.0.1/user/error-codes.html
https://pep8.readthedocs.io/en/latest/intro.html#error-codes

Is there a complete list of C9** errors?

.flake8
```diff
@@ -0,0 +1,15 @@
...
+ignore = E127, E129, E201, E202, E203, E225, E226, E231, E241, E261, E262, \
+         E265, E266, E301, E302, E303, E305, E401, E701, W391, W504, C901, \
+         F401, E228, E126, W503, E731, E211, E221, E222, E741, E251, E402, \
```
stepan-tsirkin, line 5: `F401 | module imported but unused`
I prefer **not** to ignore `F401`

stepan-tsirkin (approved):
I attest that the 'nontrivial' bugs look correct. 

So, you suggest to merge now, and in further PRs try to reduce the number of flake8 warnings to zero, right? I can 'use administrator privileges' for that so far. Or can you temporary make the `flake8` check non-mandatory for a merge?

Just address the minor issues above, and I will merge.

jaemolihm:
> So, you suggest to merge now, and in further PRs try to reduce the number of flake8 warnings to zero, right?

Yes, exactly.

> I can 'use administrator privileges' for that so far. Or can you temporary make the `flake8` check non-mandatory for a merge?

You can change the setting following https://stackoverflow.com/questions/60776412/github-actions-is-there-a-way-to-make-it-mandatory-for-pull-request-to-merge. (I do not have access to the setting.)

I also removed following checks from ignore:
E266  too many leading ‘#’ for block comment
E251  Unexpected spaces around keyword / parameter equals

.flake8
jaemolihm: Done

.flake8
jaemolihm: See https://www.flake8rules.com/.
I just removed the link from the file, because google search like "flake8 W503" easily gives the information.

PR #143, opened 2022-03-17
Fix Path when input k_nodes is a list of numpy arrays, add test

jaemolihm:
`Path` constructor was not working when `k_nodes` was a list of numpy arrays (e.g. `k_nodes=[np.array([0., 0., 0.]), np.array([0., 0., 0.5])]`).

* Fixed this bug.
* Added test that covers this case.
* Split `test_path` into four smaller tests

Before fix: 
```python3
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-7-e00ca46059d7> in <module>
      6 k_A = np.array([0.7582, 0.5000, 0.2418])
      7 
----> 8 path = wberri.Path(system, k_nodes=[k_U, k_Z, k_A], labels=["U", "Z", "A"], dk=1e-2)

~/program/wannier-berri/wannierberri/__path.py in __init__(self, system, k_list, k_nodes, length, dk, nk, labels, breaks)
     74             self.breaks=[]
     75             for start,end,l1,l2 in zip(k_nodes,k_nodes[1:],labels,labels[1:]) :
---> 76                 if None not in (start,end):
     77                 #if start is not None and end is not None
     78                     self.labels[self.K_list.shape[0]]=l1

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```


stepan-tsirkin (approved):
Well pointed!

PR #146, opened 2022-03-26
Fix symmetry Group constructor to raise for an infinite group

jaemolihm:
Previously, if a generator for an infinite group is passed to the `symmetry.Group` constructor, it did not raise but fall into an infinite loop. I fixed the constructor to raise. Added a test.

Assumes #140 is merged

Reproducer:
```python3
import numpy as np
import wannierberri.symmetry as sym
c = np.cos(0.1)
s = np.sin(0.1)
y = sym.Group([sym.Symmetry(np.array([[1, 0, 0], [0, c, s], [0, -s, c]]))])
```

### DIFF
--- a/tests/test_symmetry.py
+++ b/tests/test_symmetry.py
@@ -1,0 +1,19 @@
+"""Test symmetry objects"""
+
+import numpy as np
+import pytest
+
+import wannierberri as wberri
+import wannierberri.symmetry as sym
+
+def test_symmetry_group():
+    assert sym.Group([sym.Inversion, sym.TimeReversal]).size == 4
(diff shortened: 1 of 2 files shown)


stepan-tsirkin (approved):
I see. I think, in the old implementation the loop was not infinite, but rather too large,
it could result `~2**1000` eleemnts in the grup, before the `RuntimeError` is raised. New implementation is way better. 

btw, anyone knows what is the maximal number of elements in a magnetic spacegroup?

jaemolihm:
> btw, anyone knows what is the maximal number of elements in a magnetic spacegroup?

Wouldn't it be 96 = 48 * 2? Of course for supercells it could be arbitrarily larger.

Issue #139, opened 2022-02-11
Automatic calculation of symmetry and more symmetry-related features


jaemolihm:
Manually setting symmetries is not very difficult, although somewhat inconvenient. But in high-throughput applications, an automatic method to set symmetries will be very important.

I plan to implement the following symmetry-related features, which I think will be quite useful. I would appreciate any ideas or comments.

* Add structure information to system by `atom_positions` and `atom_magmetic_moments` field. Initialize using `system.set_structure(atom_positions, atom_magmetic_moments)`.
* Compute symmetry using [Spglib](https://spglib.github.io/spglib/python-spglib.html)
* Add method to test symmetry (e.g. compare eigenvalues, velocities, ... at symmetry-equivalent k points)
* Automatic high-symmetry k path using [seek_path](https://github.com/giovannipizzi/seekpath/). For example, band structure can be easily visualized without manually setting the k point path (e.g. `system.plot_band_structure(dk=0.1)`).

stepan-tsirkin:
> [ ]  Automatic high-symmetry k path with [seek_path](https://github.com/giovannipizzi/seekpath/). For example, band structure can be easily visualized without manually setting the k point path (e.g. `system.plot_band_structure(dk=0.1)`).

This is actually a good thing to do.  I think one should start with implementing a classmethod of Path:

```python

class Path(GridAbstract):
....
    @classmethod
    def seek(cls, system, dk):
        .... # use seekpath 
        return cls(system, labels=labels, nodes=nodes, ..)
````



Then, in  evaluate_k_path:

https://github.com/wannier-berri/wannier-berri/blob/5983c43f89c743893cea65b90d2540e05c30432b/wannierberri/evaluate_k.py#L131

if nothing is provided for the path - create it automatically

Finally, you may combine it with a call to 

https://github.com/wannier-berri/wannier-berri/blob/5983c43f89c743893cea65b90d2540e05c30432b/wannierberri/result/tabresult.py#L224

@jaemolihm, is it something you still want to do in the near future, or do you have some drafts in implementing this?

Or should I better go ahead?

Best,
Stepan.

jaemolihm:
Hello Stepan,

Please feel free to go ahead. What you described makes sense.

Note that in seekpath you need to use the function get_path_orig_cell, which does not perform standardization and keeps the "original cell".

https://seekpath.readthedocs.io/en/latest/module_guide/index.html#seekpath.getpaths.get_path_orig_cell

Best,
Jae-Mo

PR #148, opened 2022-03-26
Automatic symmetry detection using spglib

### LINKED ISSUE #144: GaAs symmetry in tests is wrong
https://github.com/wannier-berri/wannier-berri/blob/1fccb3564de3a12cacb6c1f5f5048614957de2cd/tests/create_system.py#L21-L24

C4z should be C4z * inversion.

Need to check whether this affects any results. -> It was not being used anywhere, so no problem.

jaemolihm:
* Added method `System.set_structure` that sets `positions`, `atom_numbers` (Integer numbers to distinguish species), and `magnetic_moments` (optional) attributes.
* Added method `System.set_symmetry_from_structure` that uses spglib to automatically detect symmetry of the system and sets the `symgroup` attribute.

Implements the first two steps of #139.

Assumes #140, #146 are merged.

Fixes #144.

For magnetic systems, symmetries involving time reversal (e.g. `SYM.C2x * SYM.TimeReversal`) are not detected because spglib does not support such symmetries. (For non-magnetic systems, TimeReversal is added to the generator so there is no problem.)

### TODO
* [x] Magnetic symmetry
* [x] Manual documentation in website? : Docstring is already generated.
* [x] More tests

### DIFF
--- a/wannierberri/__system.py
+++ b/wannierberri/__system.py
@@ -444,12 +444,11 @@ class System():
         """
         Set the symmetry group of the :class:`System`. Requires spglib to be installed.
         :meth:`System.set_structure` must be called in advance.
+
+        For magnetic systems, symmetries involving time reversal are not detected because
+        spglib does not support time reversal symmetry for noncollinear systems.
         """
         import spglib
-        if self.magnetic_moments is not None:
-            # Magnetic symmetry not implemented
-            # https://github.com/spglib/spglib/issues/150
-            raise NotImplementedError
 
         spglib_symmetry = spglib.get_symmetry(self.get_spglib_cell())
         symmetry_gen = []
@@ -460,13 +459,13 @@ class System():
             W = spglib_symmetry["rotations"][isym]
             Wcart = self.real_lattice.T @ W @ np.linalg.inv(self.real_lattice).T
             R = Wcart.T
-            # TODO: Time reversal information
             symmetry_gen.append(Symmetry(R))
 
         if self.magnetic_moments is None:
             symmetry_gen.append(TimeReversal)
         else:
-            print("Warning: spglib does not ")
+            print("Warning: spglib does not support time reversal symmetry for noncollinear "
+                  "systems.\nTo include such symmetries, use set_symmetry.")
 
         self.symgroup = Group(symmetry_gen, recip_lattice=self.recip_lattice, real_lattice=self.real_lattice)
 
--- a/wannierberri/__system.py
+++ b/wannierberri/__system.py
@@ -409,7 +409,7 @@ class System():
         else:
             print (f"{XX} is missing,nothing to check")
 
-    def set_structure(self, positions, atom_numbers, magnetic_moments=None):
+    def set_structure(self, positions, atom_labels, magnetic_moments=None):
         """
         Set atomic structure of the system.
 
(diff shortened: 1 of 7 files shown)


wannierberri/__system.py
```diff
@@ -409,6 +409,66 @@ def check_hermitian(self,XX):
...
 
+    def set_structure(self, positions, atom_numbers, magnetic_moments=None):
+        """
+        Set atomic structure of the system.
+
+        Parameters
+        ----------
+        positions : (num_atom, 3) array_like of float
+            Atomic positions in fractional coordinates.
+        atom_numbers: (num_atom,) list of int
```
stepan-tsirkin, line 420: would be good to provide `atom_labels` -  a list of any labels (string,number, or in general any hashable type, e.g.`['Fe','O1','O2','O1',2,3.14]` and then internally map them to numbers to pass to the `spglib`

wannierberri/__system.py
```diff
@@ -409,6 +409,66 @@ def check_hermitian(self,XX):
...
+            # 2) take transpose to go to reciprocal space.
+            W = spglib_symmetry["rotations"][isym]
+            Wcart = self.real_lattice.T @ W @ np.linalg.inv(self.real_lattice).T
+            R = Wcart.T
+            symmetry_gen.append(Symmetry(R))
+
+        if self.magnetic_moments is None:
+            symmetry_gen.append(TimeReversal)
+        else:
+            print("Warning: spglib does not support time reversal symmetry for noncollinear "
```
stepan-tsirkin, line 467: @jaemolihm , did you try testing `spglib` mith magnetic moments ? does it work?

I tried and it looks to neglect them, e.g. 

```
In [3]: spglib.get_symmetry(([[1,0,0],[0,2,0],[0,0,3]],[ [-0.1,-0.1,-0.1],[0.1,0.1,0.1]],[1,1],[-1,1]))                                                                                                                                                                                         
Out[3]: 
{'rotations': array([[[ 1,  0,  0],
         [ 0,  1,  0],
         [ 0,  0,  1]],
 
        [[-1,  0,  0],
         [ 0, -1,  0],
         [ 0,  0, -1]]], dtype=int32),
 'translations': array([[0., 0., 0.],
        [0., 0., 0.]]),
 'equivalent_atoms': array([0, 0], dtype=int32),
 'primitive_lattice': array([[1., 0., 0.],
        [0., 2., 0.],
        [0., 0., 3.]])}

In [4]: spglib.get_symmetry(([[1,0,0],[0,2,0],[0,0,3]],[ [-0.1,-0.1,-0.1],[0.1,0.1,0.1]],[1,1],[1,1]))                                                                                                                                                                                          
Out[4]: 
{'rotations': array([[[ 1,  0,  0],
         [ 0,  1,  0],
         [ 0,  0,  1]],
 
        [[-1,  0,  0],
         [ 0, -1,  0],
         [ 0,  0, -1]]], dtype=int32),
 'translations': array([[0., 0., 0.],
        [0., 0., 0.]]),
 'equivalent_atoms': array([0, 0], dtype=int32),
 'primitive_lattice': array([[1., 0., 0.],
        [0., 2., 0.],
        [0., 0., 3.]])}

```

Result is unaffected by the fourth entry of `cell`

wannierberri/__system.py
```diff
@@ -409,6 +409,66 @@ def check_hermitian(self,XX):
...
+                  "systems.\nTo include such symmetries, use set_symmetry.")
+
+        self.symgroup = Group(symmetry_gen, recip_lattice=self.recip_lattice, real_lattice=self.real_lattice)
```
stepan-tsirkin, line 470: A good way to go would be to determine all symmetries without account of magmom, and then try if the symmetries found do preserve magmom, either alone or in combination with TR

wannierberri/__system.py
jaemolihm: You are right. `spglib` does not work with magnetic moments.

stepan-tsirkin: then, maybe, for this PR just disable magnetic moments, and write in the docs that only TRS systems are allowed. implement magnetic systems later.

jaemolihm: Sorry, I misunderstood what you meant.

I think your are correct that spglib does not work for collinear magnetic moments as input. (It says it works, so maybe I am using wrong input.)

Spglib does work for noncollinear magnetic moments as input. For collinear systems, one can set the magnetic moments to e.g. [0., 0., 1.]. For example (based on your code, but changed only the magnetic moments argument to a vector:
```python3
>>> spglib.get_symmetry(([[1,0,0],[0,2,0],[0,0,3]], [ [-0.1,-0.1,-0.1], [0.1,0.1,0.1]], [1,1], [[0.,0.,1.], [0.,0.,1.]]))
{'rotations': array([[[ 1,  0,  0],
        [ 0,  1,  0],
        [ 0,  0,  1]],

       [[-1,  0,  0],
        [ 0, -1,  0],
        [ 0,  0, -1]]], dtype=int32), 'translations': array([[0., 0., 0.],
       [0., 0., 0.]]), 'equivalent_atoms': array([0, 0], dtype=int32), 'primitive_lattice': array([[1., 0., 0.],
       [0., 2., 0.],
       [0., 0., 3.]])}
>>> spglib.get_symmetry(([[1,0,0],[0,2,0],[0,0,3]], [ [-0.1,-0.1,-0.1], [0.1,0.1,0.1]], [1,1], [[0.,0.,1.], [0.,0.,-1.]]))
{'rotations': array([[[1, 0, 0],
        [0, 1, 0],
        [0, 0, 1]]], dtype=int32), 'translations': array([[0., 0., 0.]]), 'equivalent_atoms': array([0, 1], dtype=int32), 'primitive_lattice': array([[1., 0., 0.],
       [0., 2., 0.],
       [0., 0., 3.]])}
```

The limitation in this case is that spglib misses out the symmetry operations that involve time reversal (such as  time reversal * C2z). In short, spglib works but finds only a subset of the actual symmetry of the system.

The case with magnetic moments is actually being tested for Fe: see test_symmetry_spglib_Fe.

What I can do is to add an error if `magnetic_moments` is set to a number, not a vector. I will do this in this PR soon.

stepan-tsirkin: > What I can do is to raise an error if magnetic_moments is set to a list of numbers, not a list of vectors.

I think it is a way to go, not to complicate things at this point

> Or, one can convert from numbers to vectors internally

 in that case one needs to specify the quantization axis (which is not always `z`, but may have a default value `[0,0,1]`

>  The problem is that spglib does not tell you whether the given symmetry operation involves time reversal or not

Then it is (almost) useless information for us. Well, for the TR-even tensors it may be used. But I suggest to consider only the non-collinear case. Anyway, in a collinear case the spin-up and spin-down subsystems may be considered independently. 

So, just make sure magnetic moments passed to spglib are vectors not numbers, either way.

wannierberri/__system.py
jaemolihm: I agree, it seems like a good idea.

jaemolihm:
I updated the code. I used `atom_labels` instead of `atom_numbers`, and added an error if `magnetic_moments` is set to a list of numbers, not a list of 3d vectors.

wannierberri/__system.py
stepan-tsirkin: >  I was a bit reluctant to implement that because ideally that code should be in spglib.
Then go ahead and contribute to `spglib` !

PR #150, opened 2022-04-27
Fix wrong symmetry of opt_shiftcurrent

jaemolihm:
Shift current is odd under inversion, but `Iodd` was set to False. Fix this bug.

### TODO
- [x] Add test (Use symmetrized GaAs system and compare calculation w/ and w/o symmetry)


stepan-tsirkin (approved):
looks fine

PR #158, opened 2022-05-06
Organize formulas in fermiocean.py and change the units of conductivities to SI

Liu-Xiaoxiong:
Change unit of conductivities to SI.  Keep `berry_dipole` and `Der3E` dimensionless.  Corresponding conductivities with factors are `nahc` and 'Drude conductivity'. I and not sure if `Drude conductivity` is a good name.

- [ ] Organize formulas in `fermiocean.py` and add doc. 
- [ ] tests of `nahc` and `Drude conductivity` factors
- [x] Add factors of GME.
- [x] change all units to SI, and correct descriptions in `__intergrate.py`.
- [x] Change test references. (multiply by 100 except GME)


stepan-tsirkin:
Why are you doing this, if in #127 we turned towards disabling `fermiocean` and developing `calculators` 

If we rewrite all `fermiocean` quantities as a child of   [`StaticCalculator`](https://github.com/wannier-berri/wannier-berri/blob/master/wannierberri/calculators/classes.py#L11)  , then the files `__main.py`, `__integrate.py`, `__fermiocean.py`, `__evaluate.py`,  `__tabulate.py` will not be needed.

PR #163, opened 2022-05-08
separated the old API to a submodule

stepan-tsirkin:
So far it is just re-organization, to make sure that the code not using the `integrate()` and `tabulate()` rooutines explicitly does not depend on the code that is now in the `__old_API` folder.  All tests and examples should run as earlier.


jaemolihm (approved):
Looks good

PR #164, opened 2022-05-08
 allowed tabulation along a path using

stepan-tsirkin:


jaemolihm (approved):
I do not think adding `mode` field to a `Calculator` is a good idea, because a `Calculator` has nothing to do with which k points are being applied. But, I see the need for it in the construction of `TABresult` and I cannot think of a better way.

So, I approve this PR, and I will create an issue for refactoring this part.

stepan-tsirkin:
@jaemolihm , I agree.  From beginning, I did not even like packing all tabulators into `TabulatorAll()` but I do not find a good alternative so far, without loosing performance. Let's leave it to future

PR #165, opened 2022-05-09
Fermiocean to calculator with doc

Liu-Xiaoxiong:
Move functions from `fermiocean` to `calculators.static`

In the docs of the functions, all the units are in SI. But the factors haven't been changed, in order to compare with old references data. 
PS: I don't know why we are using `cm` as the length unit, but other units are in SI. If there are no special reasons, I am going to change all the units in SI (next PR).

I change some names of the functions. 
BerryDipole -> NLAHC
Der3E -> Drude(this one I am not sure if it is okay) [https://journals.aps.org/prresearch/abstract/10.1103/PhysRevResearch.2.043081](url)
Hall_spin -> Zeeman_AHC_spin
Hall_morb -> Zeeman_AHC_orb

The output of each function is dimensionless with `use_factor=False`, and with SI unit with `use_factor=True`. (after fixing all the factors)

Tests are added.

Reference paper of functions are added. (most of them)

Remove useless class `FormulaProduct_2` and `DeltaProduct` in `formula.py`. I am not sure why they are there.


stepan-tsirkin:
We use cm only for historical reasons (w90). Yes, it is more logical to switch to SI in a next PR


I guess, without factors most results are not dimensionless, but are some powers of eV and Ang. Only berry dipole is dimensionless

stepan-tsirkin:
Drude usually refers to Der2E. Der3E is nonlinear Drude

stepan-tsirkin:
Maybe AHC_Zeemann_orb and AHC_Zeemann_spin.

Liu-Xiaoxiong:
> We use cm only for historical reasons (w90). Yes, it is more logical to switch to SI in a next PR
> 
> I guess, without factors most results are not dimensionless, but are some powers of eV and Ang. Only berry dipole is dimensionless

Yes, only Berry Dipole.

>Drude usually refers to Der2E. Der3E is nonlinear Drude

NLDrude now.

>Maybe AHC_Zeemann_orb and AHC_Zeemann_spin.

Done.

Liu-Xiaoxiong:
All tests match old references.  calculator static can instead of fermiocean.py.

Hope I didn't write much typo in docs.  I checked twice, but there are some.

wannierberri/calculators/static.py
```diff
@@ -25,54 +64,552 @@ def __init__(self, fder,**kwargs):
...
+        With Fermi surface integral. Eq(8) in `Ref <https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.115.216806>`_
+        Output:
+        :math: `D_{\beta\delta} = -e^3/\hbar^2 \tau \int [dk] v_\beta \Omega_\delta f'`
+        Instruction:
+        :math: `j_\alpha = \epsilon_{\alpha\delta\gamma} \D_{\beta\delta} E_\beta E\gamma`"""
         super().__init__(**kwargs)
 
 
-class BerryDipole_FermiSea(StaticCalculator):
+class NLAHC_FermiSea(StaticCalculator):
```
stepan-tsirkin, line 452: I would still leave the calculator's name "BerryDipole_**" , it is more sound and it is nice that it is dimensionless

You may add 
```
class NLAHC_Fermisea(BerryDipole_FermiSea):
    def __init__(self,**kwargs):
        super().__init__(kwargs)
        self.factor = 
        self.comment = 
```
So, they differ only by factor and comment

Same for fermisurface.

jaemolihm:
Can you use docstrings instead of defining `self.comment`?

Also what are `if 'comment' not in dir(self)` for?


> The output of each function is dimensionless with use_factor=False, and with SI unit with use_factor=True. (after fixing all the factors)

I think you mean the results are in eV, Angstrom, etc units (the Wannier90 and WannierBerri units) when `use_factor=False`. Is this correct?

What is the difference between BerryDipole_FermiSurf and NLAHC_FermiSurf? Just the factor? If so, using `use_factor=False` with `NLAHC_FermiSurf` will give identical results with `BerryDipole_FermiSurf`, which might be  confusing.

Also, could you document the idea behind the names of the factors? For example, what does "t1", "2", and "0" mean in `factor_t1_2_0`?

stepan-tsirkin:
The idea behind `self.comment` is to pass this comment to the result, and finally wrtie to the output file, see here : https://github.com/wannier-berri/wannier-berri/blob/f3ed5620222fb9eee2e11efa29c54138c2d18144/wannierberri/__result.py#L328-L362

So, that it is clear what (and in which units) is written to that file.

Instead, docstring should contain not only the comment but also description of input parameters. Is there a nice workaround to avoid copypasting of the comment, maybe like :+1: 
```
def class MyFavouriteCalculator(....):
    __comment = "describe functionality"
    __doc__ = __comment+ "describe input parameters
    def init(self,...):
        self.comment = __comment
```
Or any better ideas?

jaemolihm:
I see. It seems generally a good idea.

Maybe we can add docstrings describing the functionality as the class docstring, and document the input parameters in the docstring for `__init__`. The python convention also says that "The class constructor should be documented in the docstring for its __init__ method." (https://peps.python.org/pep-0257/)

To access the "comment" we can use `type(calculator_instance).__doc__`.


```python3
class MyFavouriteCalculator(....):
    """describe functionality"""

    def __init__(self,...):
        """describe input parameters"""
        self.comment = __comment
```

Liu-Xiaoxiong:
> Can you use docstrings instead of defining `self.comment`?
> 
As Stepan said above. The idea is to assign variables, then print descriptions in output files. Later there should be docstrings for the webpage.

> > The output of each function is dimensionless with use_factor=False, and with SI unit with use_factor=True. (after fixing all the factors)
> 
> I think you mean the results are in eV, Angstrom, etc units (the Wannier90 and WannierBerri units) when `use_factor=False`. Is this correct?

Yes, right.

> 
> What is the difference between BerryDipole_FermiSurf and NLAHC_FermiSurf? Just the factor? If so, using `use_factor=False` with `NLAHC_FermiSurf` will give identical results with `BerryDipole_FermiSurf`, which might be confusing.
> Also what are `if 'comment' not in dir(self)` for?

These two are the same story.  The `self.comment` of BerryDipole and NLAHC are different. 
`if 'comment' not in dir(self)` can avoid BerryDiple.comment overlap NLAHC.comment.
I think we can keep them both. It can help users get what they want using default parameters. 
 
> 
> Also, could you document the idea behind the names of the factors? For example, what does "t1", "2", and "0" mean in `factor_t1_2_0`?

Okay, I will. That means the factor in the function with  t1 is the order of relaxation time. _2_0 are orders of E and B respectively.

stepan-tsirkin:
@Liu-Xiaoxiong  I added tests for `Morb` and `Morb_test`. `Morb` works, but the `Morb_test does not. Can you fix it?

wannierberri/calculators/static.py
```diff
@@ -25,41 +64,401 @@ def __init__(self, fder,**kwargs):
...
+        :math: `K^{spin}_{\alpha\mu} = e \tau \int [dk] v_\alpha s_\mu f'`
+        Instruction:
+        :math: `j_\alpha = K_{\alpha\mu} B_\mu"""
+        super().__init__(**kwargs)
+
+
+# E^1 B^0
 class AHC(StaticCalculator):
+    r"""Anomalous Hall conductivity (s^3 * A^2 / (kg * m^3) = S/m)
+        Eq(11) in `Ref <https://www.nature.com/articles/s41524-021-00498-5>`_
```
stepan-tsirkin, line 335: Can you find a more pioneering paper on AHC ?

stepan-tsirkin (approved):
amazing!

PR #168, opened 2022-05-11
Change unit to SI

Liu-Xiaoxiong:
Change unit of quantities to SI.
Including quantities in:
`fermiocean.py`
`kubo.py`
`calculator.static.py`
`calculator.dynamic.py`

Change test references.


Liu-Xiaoxiong:
codecov/patch not pass because we do not have test for `Nonlinear Drude` fder = 1 or 2. and `GME` fder = 1.

PR #169, opened 2022-05-11
All Class in Static are children of `StaticCalculator`

Liu-Xiaoxiong:
Rewrite orbital moment related Class as Child class of StaticCalculator.

Move  comments to docstrings.

Merge #168 first.


Liu-Xiaoxiong:
It works but it looks ugly on the webpage.
At least we have them.

Liu-Xiaoxiong:
@stepan-tsirkin From my side, I think I will not touch the code logic. Only add some doc for web pages. In the next days.

PR #179, opened 2022-05-15
symmetrization of magnetic system works for multi_atom unit cells.

Liu-Xiaoxiong:
Fix bug: 'symmetrization of the magnetic system only works for single-atom unit cell. '
Fix bug: ' Symmetrization of vector-matrix only works for the system have the same num_wann from each atom'

Does not affect the tutorials. We can merge later.


Liu-Xiaoxiong:
A test 'non-colinear antiferromagnetic material Mn3Sn' added.

jaemolihm:
> A test 'non-colinear antiferromagnetic material Mn3Sn' added.

Could you also add dft and wannier90 input files so that the data can be reproduced in the future?

Liu-Xiaoxiong:
> > A test 'non-colinear antiferromagnetic material Mn3Sn' added.
> 
> Could you also add dft and wannier90 input files so that the data can be reproduced in the future?

Done

wannierberri/system/sym_wann.py
```diff
@@ -309,26 +309,21 @@ def atom_rot_map(self, sym):
...
-                        print('Symmetry operator {}*T respect magnetic moment'.format(sym + 1))
-                    if sym_T + sym_only == 0:
-                        break
-
-            else:
-                sym_only = True
-                sym_T = True
+                    if sym_only or sym_T:
+                        magmom = np.round(self.wann_atom_info[i][-2], decimals=4)
+                        new_magmom = np.round(np.dot(rot_sym_glb, magmom), decimals=4)*np.linalg.det(rot_sym_glb)
```
stepan-tsirkin, line 318: Do we really need round here?

stepan-tsirkin (approved):
looks good, after my simplifications, do you agree?
remove round (above) and squash-merge

wannierberri/system/sym_wann.py
Liu-Xiaoxiong: Yes, it is not needed. (if we don't copy the result from VASP directly. And better not copy that).

PR #183, opened 2022-05-30
Fixing requrements for versions of packages

stepan-tsirkin:
I specified versions of some packages, and also removed testing of python 3.6, as some of the newer versions of packages do not support it anymore.


stepan-tsirkin:
> For packages with upper-bounded versions (numpy and protobuf), could you comment in setup.py why it is needed (with a link to github issue if possible) ? In the future, we can see if those issues are fixed and remove the upper bounds.

Done.  the numpy upper bound is that given by numba, but numpoy-1.23 is still in pre-release
With protobuf it is a very recent story, actually: https://github.com/ray-project/ray/issues/25205

PR #188, opened 2022-06-09
utils.postw90

stepan-tsirkin:
Created a wrapper, that would mimic (at least partly) the work of postw90.x, 

Mainly for testing, and also for users that are too much used to that, and for demonstration of speed, without need to change any files..

So far, only ahc is supported, added some tests

Any comments on implementation are welcome


jaemolihm:
I think you did not add some files?

stepan-tsirkin:
> I think you did not add some files?

True, now added

stepan-tsirkin:
@jaemolihm , any suggestions what to do with the failing checks?

I guess, the problem of flake8 is connected with the new [release](https://pypi.org/project/flake8/#history) as [suggested here](https://stackoverflow.com/questions/74558565/flake8-error-code-supplied-to-ignore-option-does-not-match-a-z1-30)  shal lwe modify the ignore line (how?) or restrictto an older version of flake8?

PR #191, opened 2022-06-10
Phonons

stepan-tsirkin:
Reading Phonons form QuantumEspresso AND INTERPOLATING THEM

So far implemented interpolation of phonon frequencies and phonon DOS and cumDOS. To be tested.


stepan-tsirkin:
I did some tests, looks to agree, and seems faster than the `dynmat.x` . @jaemolihm , you have some experience with phonons, right? could you try it and provide some feedback?  As you understand, it is a step towards electron-phonon interactions.

jaemolihm:
The phonon feature seems very nice. Yes, I will try it.

Just to make sure, you did not implement long-range interaction, right? So the code will work for metals and non-polar insulators, but not for polar insulators (i.e. it will not give LO-TO splitting).

Also, does the `use_ws` option works just as for electrons, just with the atomic positions instead of the Wannier centers?

stepan-tsirkin:
> Just to make sure, you did not implement long-range interaction, right

not so far. Do you have a reference to look how it should be done?

>  Also, does the use_ws option works just as for electrons, just with the atomic positions instead of the Wannier centers?

exactly

jaemolihm:
> > Just to make sure, you did not implement long-range interaction, right
> 
> not so far. Do you have a reference to look how it should be done?

The original reference is https://journals.aps.org/prb/abstract/10.1103/PhysRevB.55.10355

But the supplemental material of this recent paper is much easier to parse. https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.125.217602
(In QE, only the "DD" (dipole-dipole) term is implemented.)

For lower dimensions, one needs different formulas. For 2D, they are in appendix B of https://arxiv.org/abs/2207.10190
The 1D case is in https://arxiv.org/abs/2208.09887

wannierberri/data_K.py
```diff
@@ -311,9 +318,20 @@ def E_K_corners(self):
...
                     Ecorners[:, ix, iy, iz, :] = E[self.select_K, :][:, self.select_B]
+        Ecorners = self.phonon_freq_from_square(Ecorners)
         print_my_name_end()
+#        print ("Ecorners",Ecorners.min(),Ecorners.max(),Ecorners.mean())
         return Ecorners
 
+    def phonon_freq_from_square(self,E):
+        """takes  sqrt(|E|)*sign(E) for phonons, returrns input for electrons"""
+        if self.is_phonon:
+            e = np.sqrt(np.abs(E))*Ry_eV
```
jaemolihm, line 329: Multiply `Ry_eV**2` to `system.Ham_R` instead?

wannierberri/system/system_phonon_qe.py
```diff
@@ -0,0 +1,158 @@
...
+        cnt = 0
+        for ifile in range(1,nqirr+1):
+            fname = f"{seedname}.dyn{ifile}.xml"
+            data = untangle.parse(open(fname).read().lower()).root
+            geometry = data.geometry_info
+            number_of_atoms = int(geometry.number_of_atoms.cdata)
+            number_of_types = int(geometry.number_of_types.cdata)
+            masses_tp = np.array([float(geometry.__getattr__(f'mass_{i+1}').cdata) for i in range(number_of_types)])
+            freq = np.array( [data.frequencies_thz_cmm1.__getattr__(f"omega_{j+1}").cdata.split()[0] for j in range(3*number_of_atoms)],dtype=float  )
+            freq_cmm1 = np.array( [data.frequencies_thz_cmm1.__getattr__(f"omega_{j+1}").cdata.split()[1] for j in range(3*number_of_atoms)],dtype=float  )
```
jaemolihm, line 79: Remove variables that are not used, and read geometry info only once (from first q point)

wannierberri/system/system_phonon_qe.py
```diff
@@ -0,0 +1,158 @@
...
+                q_points.append(q)
+                cnt += 1
+                self.real_lattice, self.recip_lattice = real_recip_lattice(real_lattice=self.real_lattice)
+        self.wannier_centers_cart = self.wannier_centers_red.dot(self.real_lattice)
+        qpoints_found = np.zeros( self.mp_grid,dtype=float)
+        dynamical_matrix_q = np.zeros( tuple(self.mp_grid)+(self.number_of_phonons,)*2,dtype=complex)
+        print (dynamical_matrix_q.shape)
+        agrid = np.array(self.mp_grid)
+        print (cnt, len(q_points), len(dynamical_mat))
+        for q,dyn_mat in zip(q_points,dynamical_mat):
```
jaemolihm, line 131: Remove debug prints

jaemolihm (changes requested):
@stepan-tsirkin I fixed one bug: mass factor should be divided after ASR, not before. This effects phonon frequency where there are atoms with different masses.

After this fix, the code seems fine and working well! I left some comments.

tests/data/Si_phonons/si.dyn0
```diff
@@ -0,0 +1,5 @@
+   2   2   2
```
jaemolihm, line 1: Please add QE input files as well as pseudopotential file for the phonon test data

wannierberri/data_K.py
jaemolihm: Also `returrns`

PR #195, opened 2022-06-16
Allow running tests without gpaw and ase

### LINKED ISSUE #171: Allow testing without installing ase and gpaw

jaemolihm:
* Allow running tests with gpaw and ase installed. Tests that use gpaw will error, but others run well. (I don't have gpaw in my local computer but I wanted to run tests!)
* Make `tests/.gitignore` more general (ignore all mmn, amn, uHu, reduced_* files - they are added as compressed .tar.gz files or created on the fly)

Closes #171


stepan-tsirkin:
weird, I just tried to call the ase/gpaw only once, and this error. It does not happen on my machine locally. Any ideas?

stepan-tsirkin:
right, we require `numpy<1.23` in `setup.py`, but `pip install -U pip setuptools wheel ` in [.github/install_script.sh](https://github.com/wannier-berri/wannier-berri/tree/fixtests/.github/install_script.sh) collects `1.23`

```
Collecting numpy>=1.16
  Downloading numpy-1.23.0-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (17.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 17.1/17.1 MB 72.1 MB/s eta 0:00:00
```

How do we better handle this?

jaemolihm:
When I checked by connecting to GitHub Actions via ssh, what I found was 1) numpy 1.22 was installed and used (but gpaw complains), 2) if I force-install numpy 1.23 gpaw works (but numba would complain - I haven't checked it).

jaemolihm:
@stepan-tsirkin I could not find a way to fix the error. What about marking GPAW tests as "expected to fail" (https://docs.pytest.org/en/latest/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) and moving on? I made a commit that does so.

Importing GPAW will be tried every time, but failing import will not give an "X" in the check result.

I will also open an issue on using GPAW.

jaemolihm:
> Importing GPAW will be tried every time, but failing import will not give an "X" in the check result.

https://github.com/wannier-berri/wannier-berri/runs/7203344769?check_suite_focus=true
```
tests/test_run.py ..............................xx..                     [ 88%]
============ 92 passed, 2 xfailed, 4 warnings in 375.55s (0:06:15) =============
```

stepan-tsirkin (approved):

PR #202, opened 2022-06-30
Fix use of factor in static calculators

### LINKED ISSUE #201: factor in calculators are not respected
In the following constructor, `self.factor` gets overwritten to 1 in `super().__init__(**kwargs)`.
https://github.com/wannier-berri/wannier-berri/blob/9b06d733d6b283e3e93dff14ed2c9d3c0e681384/wannierberri/calculators/static.py#L465-L474

So, `NLAHC_FermiSurf` and `BerryDipole_FermiSurf` gives the same result. (They should differ by a factor `factor_nlahc`.)

One can fix the issue by moving all `self.factor` after `super().__init__(**kwargs)`, but then it interferes with `StaticCalculator.__init__` because of `assert hasattr(self, 'factor'), "factor not set"`.
https://github.com/wannier-berri/wannier-berri/blob/9b06d733d6b283e3e93dff14ed2c9d3c0e681384/wannierberri/calculators/static.py#L23-L29

Example:

```python3
#!/usr/bin/env python3.8
import os
import copy
import numpy as np
import matplotlib.pyplot as plt
import pickle
import wannierberri as wberri
from wannierberri import calculators as calc

seedname = "/home/jmlim/program/wannier-berri/tests/data/GaAs_Wannier90/GaAs"
system = wberri.System_w90(seedname, berry=True)

efermi = np.linspace(7, 9, 11, True)
calculators = dict(
    nlahc = calc.static.NLAHC_FermiSurf(Efermi=efermi),
    bcd = calc.static.BerryDipole_FermiSurf(Efermi=efermi),
)

result = wberri.run(
    system,
    grid = wberri.Grid(system, NK=6),
    calculators = calculators,
    print_Kpoints = False,
)

y1 = result.results["nlahc"].data
y2 = result.results["bcd"].data
print(y1 / y2)
```

jaemolihm:
* Rename `factor` in StaticCalculators to `final_factor` (to be consistent with Dynamic ones)
* Make `final_factor = 1.` the default
* Fix problem that some factors were ignored

Closes #201


jaemolihm:
Note to myself: Maybe let final_factor (constant_factor?) an argument of StaticCalculator? This will make the cases where __call__ is overriden simpler.

wannierberri/calculators/static.py
```diff
@@ -188,74 +185,54 @@ class Morb(StaticCalculator):
...
-        Omega_res = AHC(Efermi=self.Efermi, tetra=self.tetra, smoother=self.smoother,
-                use_factor=False, print_comment=False,
-                kwargs_formula=self.kwargs_formula)(data_K).mul_array(self.Efermi)
-        final_factor = -factors.eV_au / factors.bohr**2
-        if not self.use_factor:
-            final_factor = np.sign(final_factor)
-        #with use_factor=False, the factor of AHC is -1, so it is '+' below.
-        return final_factor * data_K.cell_volume * (Hplus_res + 2 * Omega_res)
+        Omega_res = self.AHC(data_K).mul_array(self.Efermi)
+        return (Hplus_res - 2 * Omega_res) * data_K.cell_volume
```
jaemolihm, line 197: I simplified this pattern (additional calculator in `__call__`) by initiating the calculator in `__init__` and storing it as an attribute.

stepan-tsirkin (approved):
Looks good, thanks

PR #203, opened 2022-07-05
Solved index issue in the FPLO Hamiltonian reading function.

philipp-eck:
One of the used indices seems to be false, the calculated distance vector (R_vec) is in turn not an integer linear combination of the Bravais vectors. This results in an assert error in line 98.
My corrent choice (iw,jw) reproduced the correct band structure.

### DIFF
--- a/wannierberri/system/system_fplo.py
+++ b/wannierberri/system/system_fplo.py
@@ -92,7 +92,7 @@ class System_fplo(System):
                             continue
                         arread = np.array(arread, dtype=float)
                         Rvec = arread[:, :3] + (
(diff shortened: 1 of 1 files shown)


jaemolihm:
Thanks for the fix! The fix indeed looks correct. (The test failure is due to gpaw installation, and I checked locally that other tests pass.)

@philipp-eck Could you confirm that the opposite choice, i.e. `self.wannier_centers_cart_auto[None, jw] - self.wannier_centers_cart_auto[None, iw])`, instead of `[iw] - [jw]`, also gives an assert error?

philipp-eck:
> 

Yes, the opposite choice results also in an assert error.

jaemolihm (approved):
Thanks for confirming.

PR #213, opened 2022-09-28
fixed spglib<2

stepan-tsirkin:
At the moment the new `spglib-2.0,x` give errors 

```
FAILED tests/test_symmetry.py::test_symmetry_spglib_Fe - AssertionError: Symmetry group size is different
```
So far let's stick to spglib<2 . Should we set a lower bound?

@Liu-Xiaoxiong , please take a note for later to figure out what they changed and how to make it compatible


Liu-Xiaoxiong:
Okay, I will check

stepan-tsirkin:
@Liu-Xiaoxiong , check, which is  lower bound for spglib version that works?

PR #216, opened 2022-12-14
Fix an edge case in DynamicCalculator

jhryoo:
In class `DynamicCalculator`, if the Fermi level is far too lower (or too higher) than energy bands so that every band is unoccupied (or occupied), then `npair == 0` and the subsequent matrix calculations emit an error because the sizes of relevant arrays are zero.

This PR fixes this bug.

### DIFF
--- a/wannierberri/calculators/dynamic.py
+++ b/wannierberri/calculators/dynamic.py
@@ -72,6 +72,9 @@ class DynamicCalculator(Calculator, abc.ABC):
                 if self.nonzero(Em, En)
             ]
             npair = len(degen_group_pairs)
+            if npair == 0:
+                continue
+
             matrix_elements = np.array(
                 [formula.trace_ln(ik, np.arange(*pair[0]), np.arange(*pair[1])) for pair in degen_group_pairs])
             factor_Efermi = np.array([self.factor_Efermi(pair[2], pair[3]) for pair in degen_group_pairs])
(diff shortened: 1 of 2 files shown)


wannierberri/calculators/dynamic.py
```diff
@@ -72,6 +72,9 @@ def __call__(self, data_K):
                 if self.nonzero(Em, En)
             ]
             npair = len(degen_group_pairs)
+            if npair == 0:
+                continue
+
```
jaemolihm, line 77: Thank you for the fix. Could you add a test case that fails before the fix and passes after the fix?

I'm not sure I understand the logic of the original code: are `Em` and `En` energies or weights? What does `Em` or `En` being zero mean?

stepan-tsirkin: `Em` and `En` are energies of bands (or degenerate groups of bands). It is not about Em and En being nonzero, it is about the contribution of the pair of bands being zero. e.g. if both states are occupied for any Fermi level considered, the pair does not contribute, and hence no need to evaluate the matrix element:

https://github.com/wannier-berri/wannier-berri/blob/27a451f7e1e7890f9256ba0a142cf5de1c18003f/wannierberri/calculators/dynamic.py#L51-L55

For other calculators this criteria may be redefined to yield False more often (e.g. due to frequency criteria)

@jhryoo , I agree, please add a test that fails before the fix.

jaemolihm: I see, thanks. I was confused by the variable name `weights` in
https://github.com/wannier-berri/wannier-berri/blob/e73b9b8ec9b26a9e37207d8cb558945fcfa3ca33/wannierberri/data_K.py#L265-L275

PR #217, opened 2022-12-14
Boosting speed in run.py

jhryoo:
This PR contains some optimizations in `run.py`:

1. When writing `K_list` into `file_Klist`, we had to write the whole `K_list` into a file at every iteration. Instead, we can append newly refined k-points only, which can significantly boost the IO performance of the code.

In the function `run`, a new flag `file_Klist_factor_changed` has been introduced. If this flag is set `True`, we append refined k-points into `file_Klist` at every iteration. Since we have to keep track of the change of weights of old k-points, a new text-formatted data file `file_Klist_factor_changed` has been introduced.

2. In the previous code, we summed the result of the calculator (e.g., the integrand of the Kubo formula) over the entire BZ at every iteration in a single processor. I modified the code so that `result_all` is updated using only the k-points that are updated in the current iteration (i.e., newly refined k-points and old k-points whose weights are changed)


jaemolihm:
Thank you for the contribution. Regarding point 1, is there any reason to keep the previous option? If not, we can just make the new option the only one.

Also, I suggest to use binary format (npy or npz) instead of txt if possible.

stepan-tsirkin:
> Thank you for the contribution. Regarding point 1, is there any reason to keep the previous option? If not, we can just make the new option the only one. Having less options is good for maintainability.

Agree.  The only drawback is that one cannot use old pickle files with a newer version, but this was never meant to. those pickles are for restart of the same calculation with the same version.

Also, if you keep only the new behaviour, there probably will be no need more tests, because we have tests that restart form a pcikle file.
 
> Also, I suggest to use binary format (npy or npz) instead of txt if possible. Maybe one can store the new information in the same pickle file for Klist.

Can one do random-access on pickle files and modify certain records only? Is it efficient for big files?

stepan-tsirkin:
I cleaned up, made the new behavior as the only one, added some tests (with two restarts) to make sure  that it is really stable.
@jhryoo , could you check if I didn't mess up anything, that is still works on your examples. 
@jaemolihm , any more comments are welcome

PS: If I do not hear from you by Feb 01, I will merge. Please feel free to ask for more time if needed.

jaemolihm:
Hi @stepan-tsirkin , I don't think I'll have the time to review the code soon. I agree with the general direction of this PR so I'm happy with merging.

PR #218, opened 2022-12-14
Change the attribute KpointBZ._max into a lazy property

jhryoo:
In class `KpointBZ`, I suggest chaing the attribute `_max` as a lazy property, since once it is calculated its value does not change. Since this attribute is calculated in every adaptive refinement iteration for all kpoints in `Klist`, caching of this attribute will make the code more efficient.


jaemolihm:
Thanks for the suggestion. Do you have a benchmark of the timings?

stepan-tsirkin (approved):
Fully agree. Definitely, this will not hurt, as far as  `_max()` returns just a small list of numbers, while to evaluate it some big arrays need to be rocessed

PR #230, opened 2023-02-16
Symmetric Tetrahedron method for trigonal lattices

stepan-tsirkin:
The standard division of space into tetrahedrons generally does not respect any symmetries.
This PR implements the division of space in a trigonal (3-fold symmetric) way.  Basically, a hexagonal prism is built around each point. and further divided into tetrahedra (again, one vortex is in the center of the prism)

Not sure, if it is correct when the refinement is done, because a hexagon cannot be broken into hexagons.


stepan-tsirkin:
a better and more universal solution is offered in #233  . So, I am closing this

PR #245, opened 2023-05-17
rearrangements in sym_wann

stepan-tsirkin:
some rearrangements to make code more laconic

### DIFF
--- a/wannierberri/system/sym_wann.py
+++ b/wannierberri/system/sym_wann.py
@@ -187,28 +187,9 @@ class SymWann():
         if self.Inv:
             print('====================\nSystem have inversion symmetry\n====================')
 
-        if self.soc:
-            if self.DFT_code.lower() == 'vasp':
-                pass
-            elif self.DFT_code.lower() in ['qe', 'quantum_espresso', 'espresso']:
-
-                def spin_range(Mat_in):
-                    Mat_out = np.zeros(np.shape(Mat_in), dtype=complex)
-                    Mat_out[0:self.num_wann // 2, 0:self.num_wann // 2, :] = Mat_in[0:self.num_wann:2,
-                                                                                    0:self.num_wann:2, :]
-                    Mat_out[self.num_wann // 2:self.num_wann, 0:self.num_wann // 2, :] = Mat_in[1:self.num_wann:2,
-                                                                                                0:self.num_wann:2, :]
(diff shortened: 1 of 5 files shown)


Liu-Xiaoxiong:
The parties are useless here. So I will remove it.

wannierberri/system/sym_wann.py
```diff
@@ -69,23 +70,29 @@ def __init__(
...
-        self.proj = proj
-        self.possible_matrix_list = ['AA', 'SS', 'BB', 'CC']  #['AA','BB','CC','SS','SA','SHA','SR','SH','SHR']
-        self.matrix_list = {k: v for k, v in XX_R.items() if k in self.possible_matrix_list}
+        self.possible_matrix_list = ['Ham','AA', 'SS', 'BB', 'CC']  #['AA','BB','CC','SS','SA','SHA','SR','SH','SHR']
+        self.matrix_list = XX_R
+        for k in XX_R:
+            if k not in self.possible_matrix_list:
+                print (f"WARNING: symmetrization of matrix {k} is not implemented yet, so it will not be symmetrized, but passed as it. Use on your own risk")
+        # This is confusing, actually the I-odd vectors have "+1" here, because the minus is already in the rotation matrix
+        # but Ham is a scalar, so +1
```
Liu-Xiaoxiong, line 85: I will cancel them

Liu-Xiaoxiong (approved):

stepan-tsirkin:
> The parties are useless here. So I will remove it. 

I do not think that you are right. Parities **are** needed here. At least because spin and position matrix element may transformed differently under the same symmetry operation, e.g in case of inversion or mirror.

Note, now the tests fail heavily with large errors. Before your commit, only the sparse tests gave an error slightly above the threshold, but still small 

For me a more clear solution is to make the cartesian rotation matrix represent only the pure rotation ( so that it's determinant equals +1) and then multiply by parity_I if the operation includes inversion.

Liu-Xiaoxiong:
The parity should be included in the symmetry operation. The different matrices respect different parity by themselves, and don't need additional parity. For the testing, I think I should check the old test of orbital magnetic moment is right or wrong. especially the part from CC_R. Let's see, maybe the test is wrong.
________________________________

stepan-tsirkin:
Anyway, the goal of this PR was to refactor code without changing any result, so that the old tests pass. If you think that the code was wrong, open another PR that only fixes the bug and updates the tests. Thus, the changes in code will be local and easier to judge if they are correct or not.

So far, please remove your latest commit from this PR

PR #249, opened 2023-06-02
translational invariance for off-diagonal position matrix elements

stepan-tsirkin:
I,


stepan-tsirkin:
Reopened as  #250 , closing here

PR #250, opened 2023-06-05
Transl_inv_offdiag  - new

stepan-tsirkin:
Because #249 is so much behind master that it is hard to merge, I opened this ne PR by just copypasting corresponding parts (implemented by @jaemolihm )

Also, the corresponding test in #249 uses the reference data from the present PR, so they work in the same way


stepan-tsirkin:
completed in #308

stepan-tsirkin:
completed in #308

PR #273, opened 2023-09-07
transl_inv_OSD

scarpoz:
The aim of this branch is to settle the real-space matrix elements that are required to compute the spatially-dispersive conductivity tensor. These are computed following Jae-Mo's approach and working in phase convention I. The R-dependent phase factors are included after ws_dist() is applied, which solves symmetry inconsistencies previously found on the Berry curvature and the orbital magnetic moment.


wannierberri/system/system_w90.py
```diff
@@ -74,10 +80,15 @@ def __init__(
...
         self.wannier_centers_cart_auto = chk.wannier_centers
 
+        # Necessary ab initio matrices
         eig = EIG(seedname)
         if self.need_R_any(['AA','BB']):
             mmn = MMN(seedname, npar=npar)
-
+        if self.need_R_any(['CC']):
+            uhu = UHU(seedname)
+        if self.need_R_any(['CC','FF','FF_tot']):
```
stepan-tsirkin, line 89: Do we need uiu, if we need only the `CC` ? I guess, it is not needed, e.g. in the case of only orbital moment

stepan-tsirkin:
@scarpoz , you need to manually resolve conflicts. mainly, in data_K, note, there are now two classes : `data_K_R` (for wannier and TB models) and `data_K_k` (for k.p models) . and also their common parent _Data_K .   All the `transl_inv` stuff is relevant only for `data_K_R`.

stepan-tsirkin:
@scarpoz  in #285  I implemented tests for the data stored in systems. You may use those tests to make similar ones for systems with translational invariance.

Issue #274, opened 2023-09-08
Incorrect Spin Hall Conductivity Calculation using WannierBerri and VASP


ma455173220:
Hi All,

I am currently working on utilizing Wannierberri to calculate the spin Hall conductivity of various materials. As part of my preliminary efforts, I conducted benchmarking by performing calculations on Pt using VASP. Although the band structure closely aligns with experimental data near the Fermi level, as convincingly demonstrated in the attached document "Band structure - wannier90_VASP_W90_cmp.pdf " under VASP folder, I have encountered an anomaly. The computed spin Hall conductivity appears to be much smaller and inaccurate (data in WannierBerri-SHC.xlsx under VASP folder).
 
To further investigate this issue, I calculated the spin Hall conductivity using the "wannier90_hr.dat" file and Wanniertools. The calculated results are consistent with the reported data (data in WannierTools-SHC.xlsx under VASP folder). A direct comparison between the results obtained from WannierTools and WannierBerri can be found in "Compare WannierTools and WannierBerri.png" under VASP folder:
![Compare WannierTools and WannierBerri](https://github.com/wannier-berri/wannier-berri/assets/42956329/5b8da070-8d7b-4382-ae44-b9f08c2eed6c)

I also performed calculations using a significantly denser k-point mesh (15x15x15). The data obtained from WannierBerri appears to be more accurate compared to the results obtained with a smaller k-point mesh. However, it's worth noting that the magnitude of the results at the Fermi level is still considerably lower:
![Compare WannierTools and WannierBerri - kmesh-15x15x15](https://github.com/wannier-berri/wannier-berri/assets/42956329/7aa0aa53-cc87-4d3a-81e9-1b320cbf00e2)

Additionally, I followed the example provided by Wannierberri, which utilizes QE, and in this case, the spin Hall conductivity was correctly computed for both WannierTools and WannierBerri (data can be found under QE folder). Therefore, I suspect that the discrepancy in the spin Hall conductivity value, specifically when using Wannierberri in conjunction with VASP, may be attributed to the "wannier90.spn" file.
 
If it is not too much trouble, would you be willing to review my input settings to identify any potential issues? All the input and output files are attached. (Notes: In WannierTools, the unit of the Spin Hall Conductivity (SHC) is (ħ/e) S/cm, which means you should divide WannierBerri's results by 100 to obtain the comparable values. Additionally, please note that the data from WannierTools has already been adjusted so that the Fermi level is shifted to 0.)

[...]

PR #279, opened 2023-09-21
Allow to manually set spin pairs

stepan-tsirkin:
As noted in #274 by @ma455173220 WannierTools by @quanshengwu  is able to calculate spin Hall without .spn file. This is done using an approximation that all WFs are pure states of Sz  

https://github.com/quanshengwu/wannier_tools/blob/1cd2d091bd6363b9f43c71c958c5eac8ef45b08b/src/Boltz_transport_anomalous.f90#L575-L586

This pull request implements a tight-binding approximation to spin Hall conductivity, analagous to the approximation made in WanierTools 

https://github.com/quanshengwu/wannier_tools/blob/1cd2d091bd6363b9f43c71c958c5eac8ef45b08b/src/Boltz_transport_anomalous.f90#L646-L654

Calculations may be executed with something like:

``` python
import wannierberri as wb
system = wb.system.System_w90(... ,  spin=False ) # SHCqiao=False, SHCryoo-=False by default

parameters_shc = dict(
        Efermi=np.array([17,18]), omega=np.arange(0.0, 7.1, 1.0), smr_fixed_width=0.20, smr_type="Gaussian"
        SHC_type="simple",kwargs_formula={"external_terms":True}

)
calculators = {'SHCsimple' : wberri.calculators.dynamic.SHC( **parameters_shc) }
wb.run(system, grid, calculators=calculators, ...  )

```

The output is to be tested


stepan-tsirkin:
I tried using this [new tutorial](https://github.com/wannier-berri/WannierBerri-tutorial/blob/c9c4381284a63833e6eb87586dccc3174e87cd11/tutorials/3_spin_hall/tutorial_spin_hall_simple.ipynb)  and got the following:

image

Looks like SHCsimple works nice for this case, if we retain the external terms. 
Aslo set_spin does not make to much difference here.

**Note**: maximal localization should be avoided with `set_spin_from_code`

ma455173220:
Hi Dr. Stepan,

I've also carried out tests on Pt using the new SHCsimple code. The results appear to be quite similar to those obtained with WannierTools. I suspect that the minor discrepancies might also be linked to the broadening parameters employed in WannierTools.

I have attached three screenshots for your reference:

- One with 'use_ws=false' and 'external_terms=false.'
- One with 'use_ws=true' and 'external_terms=false.'
- One with 'use_ws=true' and 'external_terms=true.'

It appears that for the Pt system, 'use_ws' doesn't seem to have any noticeable effect, and 'external_terms' has only a minor impact.

Regards,
Hongyang

Pt_use ws-true_external-true
Pt_use ws-true_external-false
Pt_use ws-false_external-false

stepan-tsirkin:
> It appears that for the Pt system, 'use_ws' doesn't seem to have any noticeable effect,

Right. When all WFs are at the same site of the unit cell (or close to it), use_ws has no effect at all.

PR #307, opened 2024-01-26
Disable symmetrization for convention II

sjhong6230:
The current implementation of symmetrization in Wannierberri is wrong for conventions I and II. Since the symmetrization formula is much simpler in convention I, it is desirable to revise the symmetrization for convention I and disable it for convention II.

Below is the derivation of the right symmetrization formula for convention I.

[convention_I.pdf](https://github.com/wannier-berri/wannier-berri/files/14059350/convention_I.pdf)


jaemolihm:
Since only AA is more often than BB or CC, and CC is useless without AA, I believe one can just disable these two cases: (1) use_wcc_phase=true, and (2) symmetrization of BB and/or CC.

If you expect to push a fix for all cases soon, than that would also be fine.

stepan-tsirkin:
> Since only AA is more often than BB or CC, and CC is useless without AA, I believe one can just disable (i.e. throw) these two cases: (1) use_wcc_phase=true, and (2) symmetrization of BB and/or CC.

You probably  meant disabling the combination "(use_wcc_phase=False) and ( symmetrization of BB and/or CC)." This might also be a fix. 

However, in recent years we found so many cases that convention I makes life easier, and no case when convention II is superior.  In turn, keeping the two conventions complicates the code. Therefore, there is a strategic plan to remove convention II at all from the code. 

Do you have any reason that makes it really necessary to keep convention II?  

Actually @ivosouza0 found notes from David Vanderbilt from 2007, when they discussed switching Wannier90 to convention I, but that still did not happen.
> 
> If you expect to push a fix for all cases soon, then that would also be fine.  

Therefore, a fix for convention II is not really needed.

jaemolihm:
Hi @stepan-tsirkin , I totally agree that we can deprecate convention II.

The only reason I can think of where keeping convention II is helpful is that it helps one to spot issues in the formalism (e.g. https://journals.aps.org/prb/abstract/10.1103/PhysRevB.103.247101), but keeping it incurs much more bugs than those it fixes.

sjhong6230:
While computing the orbital magnetization of bcc Fe using the conventional cell (not the bcc primitive cell), I think there is a bug in this code. I used the latest revised version of this branch, and with `transl_inv=False`. Without symmetrization, it gives the right result, that is, almost twice the value computed with the primitive cell. However, with symmetrization, the result is about 1/10 of the unsymmetrized value. I think we should merge this branch after resolving this issue.

stepan-tsirkin:
> While computing the orbital magnetization of bcc Fe using the conventional cell (not the bcc primitive cell), 

Interesting. I actually never used this symmetrization with non-primitive cells, so cannot tell you if this case is accounted properly. In principle it should work, but surprizes may happen. 

> I think there is a bug in this code. I used the latest revised version of this branch, and with `transl_inv=False`.

What about transl_inv=True?  (i.e. for diagonal components) . Because translational invariance if of higher importance for supercells. (although it should not make a difference of a factor of 10)

> Without symmetrization, it gives the right result, that is, almost twice the value computed with the primitive cell. However, with symmetrization, the result is about 1/10 of the unsymmetrized value. I think we should merge this branch after resolving this issue.

In any case, could you share your example somehow?

sjhong6230:
> > While computing the orbital magnetization of bcc Fe using the conventional cell (not the bcc primitive cell),
> 
> Interesting. I actually never used this symmetrization with non-primitive cells, so cannot tell you if this case is accounted properly. In principle it should work, but surprizes may happen.
> 
> > I think there is a bug in this code. I used the latest revised version of this branch, and with `transl_inv=False`.
> 
> What about transl_inv=True? (i.e. for diagonal components) . Because translational invariance if of higher importance for supercells. (although it should not make a difference of a factor of 10)

`transl_inv=True` does not change the results significantly.

> > Without symmetrization, it gives the right result, that is, almost twice the value computed with the primitive cell. However, with symmetrization, the result is about 1/10 of the unsymmetrized value. I think we should merge this branch after resolving this issue.
> 
> Also, is the problem after symmetrization only for Morb, or already for simpler quantities, like AHC, or simply DOS?

It also happens for AHC and DOS, too.

> In any case, could you share your example somehow?

I attached the QE, W90, WB input, pseudopotential, and the final results. 
[Fe.tar.gz](https://github.com/wannier-berri/wannier-berri/files/14118856/Fe.tar.gz)

stepan-tsirkin:
> > In any case, could you share your example somehow?
> 
> I attached the QE, W90, WB input, pseudopotential, and the final results. [Fe.tar.gz](https://github.com/wannier-berri/wannier-berri/files/14118856/Fe.tar.gz)

A few suggestions on calculations:

1.  Calculating only one Fermi level does not give confidence. It may be just a random peak, especially when dealing with a Fermi-surface property like DOS without tetrahedron method. Better do a range of Fermi levels and get a curve - it will take almost the same time as one Fermi level. 
2.  doing a run() for every calculator separately is inefficient - why not doing them all together in one run()
3. When you do not use k-space symmetries (you did not do `system.set_symmetry(...)`) adaptive iterations may even make results worse. Because at any particular iteration one k-point of a star may be refined, and other not.

Still not sure if the issue is in those points, or elsewhere, I will try to redo the calculations. But if DOS is off, it means that the Hamiltonian is messed, and that has nothing to do with phase conventions.

stepan-tsirkin:
@sjhong6230 , I repeated your calculations and observed the following:

1. The projections are defined as follows :  Screenshot from 2024-02-01 12-50-10
i.e. t2g goes before sp3d2
2. The Wanier centers after disentanglement are : 
Screenshot from 2024-02-01 12-45-20
WF 1-2 are s states, 3-8 are p-states, and WF 9-20 look like sp3d2 - they are displaced from the origin in directions of one of the axes. Next, WFs 21-26 look lije t2f - they have center on the atom. For the other atom the situation repeat.
3. If I do symmetrization, (as you did) giving the projections to wannierberri in the same order as in the win file, the HAmiltonian is messed, as we can see in the DOS - the bands are strongly displaced. However, If I give the projections in a different order (the order that I observe from Wannier centers) - the DOS looks reasonable.

This all looks weird, but it is not connected to the present PR, so it may be opened as a separate issue. 

All these is done with following versions of software: 

- wannierberri-0.15.0
- PWSCF v.6.7MaX
- Wannier90: 3.1.0
-

stepan-tsirkin:
@sjhong6230 , looks like it is a general issue with the order of projections in the win file: if you give them in one line, they are ordered by the l number, and as far as sp3d2 has l=-5, it comes first (see user guide)

I opened an [issue](https://github.com/wannier-developers/wannier90/issues/463) on wanier90 repository to hear some comments from the developers. But I think, we should clarify this point already in the docstring of `symmetrize` function.

sjhong6230:
>I opened an https://github.com/wannier-developers/wannier90/issues/463 on wanier90 repository to hear some comments from the developers. But I think, we should clarify this point already in the docstring of symmetrize function.

@stepan-tsirkin I read your comments and agree with the idea that we should document the ordering issue in the symmetrization routine.

Also, I am computing other quantities (AHC and Morb) with the right ordering and a range of fermi levels.

> 1. Calculating only one Fermi level does not give confidence. It may be just a random peak, especially when dealing with a Fermi-surface property like DOS without tetrahedron method. Better do a range of Fermi levels and get a curve - it will take almost the same time as one Fermi level.
> 2. doing a run() for every calculator separately is inefficient - why not doing them all together in one run()
> 3. When you do not use k-space symmetries (you did not do system.set_symmetry(...)) adaptive iterations may even make results worse. Because at any particular iteration one k-point of a star may be refined, and other not.

Lastly, thanks for the general comments about using wannierberri for computation. As my experience in the software is not that long, I was not aware of these issues.

PR #308, opened 2024-02-01
Spatially-dispersive conductivity tensor

scarpoz:
Implementation of the spatially-dispersive conductivity tensor. The necessary real-space matrix elements are obtained using Jae-Mo's approach to preserve translational invariance.


sjhong6230:
@scarpoz Thanks for the implementation. I used your code to compute the orbital magnetization of bcc Fe.

However, I think there is a slight mistake in the formula. The computed values of orbital magnetization do not converge to the same value for `transl_inv_JM=False` and `transl_inv_JM=True`.

Below are the results of my computation for both conventions I and II.
[convention_I_z.pdf](https://github.com/wannier-berri/wannier-berri/files/14173950/convention_I_z.pdf)
[convention_II_z.pdf](https://github.com/wannier-berri/wannier-berri/files/14173951/convention_II_z.pdf)

The $\alpha$ is a parameter that fixes the freedom of choosing the internally dividing points for the translation invariant formalism of BB. While this freedom is fixed to the center point for AA and CC due to Hermiticity, there is no such condition for BB. Thus, I checked the effect of varying $\alpha$. The $\alpha=1/2$ corresponds to your implementation.

I think the following lines are the cause of this problem.
https://github.com/scarpoz/SDCT/blob/1aed364fb4d3b76ce14a135faa8eaa0528260d73/wannierberri/system/system_w90.py#L313-L318

Revising these lines to the following resolved this problem.
https://github.com/sjhong6230/wannier-berri-sym/blob/ce688e71e6a3d66ea0f99f43a0eb0c0ac83427bd/wannierberri/system/system_w90.py#L314-L319

The results with the revised codes are the following. It converges to the same value.
[convention_I_z.pdf](https://github.com/wannier-berri/wannier-berri/files/14174024/convention_I_z.pdf)
[convention_II_z.pdf](https://github.com/wannier-berri/wannier-berri/files/14174025/convention_II_z.pdf)

stepan-tsirkin:
@sjhong6230 , well spotted. Could you commit those changes to this branch?

sjhong6230:
> @sjhong6230 , well spotted. Could you commit those changes to this branch?

Yes, since I do not have write access to this branch, I just made a pull request in @scarpoz's fork.
Please let me know if it is not the best way to do.

stepan-tsirkin:
I will squash-merge now, after the tests complete. This will fix the state of the code at this point. 

I have covered almost all changes by test. Further, we may work on improvements, optimization, etc, and we can be sure that changes do not affect the results. If bugs are found. those also may be fixed.

PR #316, opened 2024-02-19
Symmetrize SHC, OO, GG

stepan-tsirkin:
Implemented symmetrization of SHC matrix elements, (SHA, SA etc)

I assume (have no proof yet) that all matrix elements transform in convention I in a simple way (like others)

One test passes for Fe, but more testing is needed

Weird thing noticed : changing the parity_TR for the SHC matrix elements does not make the test fail

OO and GG are not checked yet


wannierberri/system/sym_wann.py
```diff
@@ -100,16 +107,20 @@ def __init__(
...
         self.parity_TR = {
             'Ham': 1,
             'AA': 1,
             '_WCC': 1,
             'BB': 1,
             'CC': -1,
-            'SS': -1
-        }  # {'AA':1,'BB':1,'CC':1,'SS':-1,'SA':1,'SHA':1,'SR':1,'SH':1,'SHR':1}
+            'SS': -1,
+            'SA': -1, 'SHA': -1, 'SR': -1, 'SH': -1, 'SHR': -1,
```
stepan-tsirkin, line 121: .. if I change any of the signs in this line, the test still passes. So, we need more checks on this.

wannierberri/system/sym_wann.py
```diff
@@ -100,16 +107,20 @@ def __init__(
             '_WCC': 1,
             'BB': 1,
             'CC': -1,
-            'SS': -1
-        }  # {'AA':1,'BB':1,'CC':1,'SS':-1,'SA':1,'SHA':1,'SR':1,'SH':1,'SHR':1}
+            'SS': -1,
+            'SA': 1, 'SHA': 1, 'SR': 1, 'SH': 1, 'SHR': 1,
```
stepan-tsirkin, line 111: If I change any of the signs in this line, the test fails. Which suggests that the signs are correct, but ...

stepan-tsirkin:
Hi all,
@andreaurru1992 @sjhong6230 @jaemolihm  @manxkim @ivosouza0 

I made the changes that allow to generally symmetrize the matrix elements for SHC and OSD.

Now what is left - to make sure it works correctly.

First, I assume (have no proof yet) that all matrix elements transform in convention I in a simple way (like others), i.e. no additional terms needed. Could any of you take a lead to write a note proving that?

This symmetrization works for a system like:

``` python
System_w90( ... , transl_inv=True, use_wcc_phase=True, wcc_phase_fin_diff=True ) # i.e. the default values now
```

Now I evaluate some integrals with and without k-space symmetrization and see if there any difference. looks like there is none for SHC (and other properties which were tested before) 

see above about the parity_TR issue

I did not check any of the properties derived from OO and GG matrix elements so far

It would be great if you do some more testing of this symetrization and tell me if it works correctly. 

BTW, I did the implementation of ` wcc_phase_fin_diff=True` for SHC matrix elements just by analogy with the others (introducing phases associated with the b vectors) in the previous PR (already merged). It would be good if you test if the results of this behaviour are as expected.

sjhong6230:
I have a question about the following lines.

https://github.com/wannier-berri/wannier-berri/blob/3316371157e5f3b33f63065e89db656edb809ce9/wannierberri/system/system_w90.py#L297-L321

`transl_inv_JM=true`  assumes convention I, and thus, needs recentering terms. However, I am curious why there are no recentering terms for GG and SHC matrix elements.

stepan-tsirkin:
@sjhong6230 
For the SHC nobody figured out what those recentering terms look like. Therefore transl_inv_JM will not work  for them, therefore in these lines

https://github.com/wannier-berri/wannier-berri/blob/3316371157e5f3b33f63065e89db656edb809ce9/wannierberri/system/system_w90.py#L91-L96

an error will be raised if SHC is needed. 

For GG, as I understand,  @scarpoz somehow figured out that those recentering terms are zero

At least, that was in his original implementation:

https://github.com/scarpoz/SDCT/blob/f57587e6b3d784ab487df665d8bc193e9efd12dc/wannierberri/system/system_w90.py#L414-L420

It does not look right to you?

sjhong6230:
I just uploaded a PR for the symmetrization of SHC. This also includes the translation invariance and convention conversion of SHC matrix elements.

#324

PR #324, opened 2024-04-24
Translation Invariant Computation and Symmetrization of SHC

sjhong6230:
In this pull request, I implemented the translation invariance and symmetrization of SHC.

The formulas for SA and SHA matrix elements are analogous to the BB matrix elements.

For simplicity, only the `SHCryoo` case was considered.

Below are the test results for translation invariance. (SHC of GaAs)

- [no_transl.pdf](https://github.com/wannier-berri/wannier-berri/files/15091264/no_transl.pdf)
- [transl.pdf](https://github.com/wannier-berri/wannier-berri/files/15091269/transl.pdf)

The plots were magnified for visibility. For `no_transl.pdf`, the result changes before and after translating the whole system. For `transl.pdf`, that does not happen, and convergence is faster. (More evident in the imaginary part.)

Below are the test results for symmetrization. (SHC of GaAs)

- [sym.pdf](https://github.com/wannier-berri/wannier-berri/files/15091298/sym.pdf)

We compared the $xyz$ and $yzx$ components of the SHC. After symmetrization, the difference becomes almost zero.


stepan-tsirkin (dismissed):
Looks good. Thank you!

sjhong6230:
Thanks! I added the GaAs test for the translation invariance of SHC.

Also, the Fe and GaAs_random test for the symmetrization was included.

One weird thing: While the Fe SHC test passes, the GaAs_random test fails.

wannierberri/system/sym_wann.py
```diff
@@ -93,14 +93,20 @@ def __init__(
             'AA': 1,
             'BB': 1,
             'CC': -1,
-            'SS': -1
+            'SS': -1,
+            'SH': -1,
+            'SA': 1,
+            'SHA': 1
```
stepan-tsirkin, line 99: Are you sure about these values for SA, SHA? 
The comment line (by @Liu-Xiaoxiong ) says `the I-odd vectors have "+1" here`, but that's only about vectors, not scalars (Ham) or rank-2 tensors.

I would formulate it as, the sgn corresponds to the number of pseudovectors in the tensor. I mean, SS is spin, so it is pseudovector, CC has vector product, so i is pseudo vetor. SA and SHA are "direct product of spin and position", so it sghould be "-1" . If we had a direct product of two spins, or vector product of spin and A, that would be "+1".  Do you agree?
Note, in #316  put "-1" for those matrices, and it seems to work.

sjhong6230:
Thanks for the comment. I understand the issue.

I revised the parity factors, and also the mp_grid in #326. Both are needed to pass the test.

Since mp_grid is changed, I also changed some test reference files for `test_system_random_GaAs_ws` and `test_system_random_GaAs_ws_sym`.

PR #343, opened 2024-07-24
Order bk vectors consistently for all k-points

stepan-tsirkin:
Now during the read of  mmn file, the bk vectors (pointing to nearest neighbors in k-grid) are ordered consistently for all k-points.
This order is also ascending in the lengths of b-vectors (when they differ, within the group of b-vectors of same length they are ordered arbitrary, but consistently for all k-points)
During the read of uHu uIu, sHu, sIu files the ordering is done accordingly

Now bk_cart and bk_latt have dimensions (NNB, 3)  , and weights are NNB - same for all k-points

Thus,  no need for bk)cart_unique, bk_latt_unique

Accordingly, the code for finite-difference is simplified.

### DIFF
--- a/tests/test_formatted.py
+++ b/tests/test_formatted.py
@@ -61,12 +61,12 @@ def generate_formatted_files(create_files_GaAs_W90):
 
 def test_formatted_uXu(generate_formatted_files):
     data_dir = generate_formatted_files
-    uHu_unformatted = UHU(os.path.join(data_dir, "GaAs"))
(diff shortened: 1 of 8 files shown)


stepan-tsirkin:
Ok, I realised that because we can save files at different points as w90-style or npz, it is hard to keep track of consistency or the b-reordering. uin mmn, uHu, uIu, sHu , etc 

So, this reordering may create more confusion and unpredictable behaviours, rather then simplify any calculation. It is better to keep the ordering as it came fro mpw2wannier90.

Closing this PR

PR #361, opened 2024-11-10
numpy version

Liu-Xiaoxiong:
fixed numpy version request in setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,8 +29,8 @@ setuptools.setup(
      long_description_content_type="text/markdown",
      python_requires='>=3.7',
      install_requires=[
-                        'numpy>=2.0', # min 1.24 required by scipy.io
-                        'scipy>=1.13',
+                        'numpy>=1.20', # min 1.24 required by scipy.io
+                        #'scipy>=1.13',
                         'colorama',
                         'termcolor',
                         'numba>=0.55.2',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -29,7 +29,7 @@
      long_description_content_type="text/markdown",
      python_requires='>=3.7',
      install_requires=[
-                        'numpy>=2.0', # min 1.24 required by scipy.io
+                        'numpy>=1.20', # min 1.24 required by scipy.io
```
stepan-tsirkin, line 32: Why exactly 1.20? 

At least Scipy requires 1.23.5 : 

https://github.com/scipy/scipy/blob/4d2f88a166ab7be1f4b16ad3aec7cde9ecdb1fef/pyproject.toml#L48-L50

and why at all not switch generally to 2.x ?

Liu-Xiaoxiong: Aha, Sorry.
My old 'pip' and me all don't know Numpy have a version 2.   ...  
I thought it was a typo. I should update myself as well as my pip. 

And in Numpy 1. It looks only 1.23.5 works.  1.24 not work.

Okay, close it .

By the way, nice wannierization part.  I am trying to test some noncolinear antiferro-magnet cases. 

Am I the first user?

PR #415, opened 2025-07-03
spin doubling

stepan-tsirkin:


phibeck:
I tested this and confirm that it works as intended. :+1:

PR #422, opened 2025-07-11
set_soc for w90data from gpaw-like

stepan-tsirkin:


stepan-tsirkin:
For now I am closing this PR. A better solution to include SOC with GPAW is introduced in #424

stepan-tsirkin:
For now I am closing this PR. A better solution to include SOC with GPAW is introduced in https://github.com/wannier-berri/wannier-berri/pull/424

PR #488, opened 2026-02-12
fix wrong return type in VoidSymmetrizer

Joh11:
Calling `wannierberri.wannierise.wannierise` was broken on master without a symmetrizer, because `select_full_blocks` for the dummy `VoidSymmetrizer` returned a tuple with None, while the symmetrizer only returns the `selected_bands_pool`. 

I don't know the code base so I'm not sure if I am missing something important, but this seems to work.


stepan-tsirkin (approved):
That's correct, thanks.

PR #524, opened 2026-03-27
Add spin Berry dipole Fermi-surface and Fermi-sea calculators

bowenphys:
This PR adds spin Berry dipole support in both Fermi-surface and Fermi-sea forms. It introduces `VelSpinOmega`, `DerSpinOmegaSimple`, `SpinBerryDipole_FermiSurf`, and `SpinBerryDipole_FermiSea`. For now the Fermi-sea implementation is only enabled for `spin_current_type="simple"` with `external_terms=False`, which is the case validated here.

I also added a small regression test in `tests/test_spin_bcd.py` to check Fermi-surface / Fermi-sea consistency on a minimal spinful TB model and to make sure unsupported Fermi-sea settings fail explicitly.


wannierberri/formula/covariant.py
```diff
@@ -780,6 +780,104 @@ def ln(self, ik, inn, out):
...
+        super().__init__(data_K, external_terms=external_terms, **parameters)
+
+        if spin_current_type != "simple" or external_terms:
+            raise NotImplementedError(
+                "DerSpinOmegaSimple currently supports only "
+                "spin_current_type='simple' with external_terms=False"
+            )
+
+        self.ndim = 4
+        self.transformTR = transform_ident
```
stepan-tsirkin: Are you sure about it?
Because, for SpinOmega both transforms are `transform_ident`
https://github.com/wannier-berri/wannier-berri/blob/f9a947d6a77629295b6fff3887e74fa7056f0818/wannierberri/formula/covariant.py#L761-L762

And derivative is odd under both, so typically it "flips" the parity of both the transforms.  Did you check your implementation against pointgroup symmetries?

wannierberri/formula/covariant.py
```diff
@@ -780,6 +780,104 @@ def ln(self, ik, inn, out):
         raise NotImplementedError()
 
 
+def _covariant_generalized_derivative(full, d_full, D):
```
stepan-tsirkin: This is very similar to the 

class Matrix_GenDer_ln(Formula_ln)

https://github.com/wannier-berri/wannier-berri/blob/f9a947d6a77629295b6fff3887e74fa7056f0818/wannierberri/formula/formula.py#L95-L96

Is there a reason not to use that class here?

wannierberri/formula/covariant.py
```diff
@@ -780,6 +780,104 @@ def ln(self, ik, inn, out):
...
+        S = data_K.Xbar("SS")
+        dS = data_K.Xbar("SS", 1)
+        V = Velocity(data_K, external_terms=False).matrix
+        dV = data_K.Xbar("Ham", 2)
+        D = np.array(data_K.D_H, copy=False)
+
+        dEinv = np.array(data_K.dEig_inv, copy=True)
+        nw = dEinv.shape[1]
+        diag = np.arange(nw)
+        for ik in range(dEinv.shape[0]):
```
stepan-tsirkin: This is not needed, because this array is already zero, where the  energies are degenerate. (including the diagonal), see

https://github.com/wannier-berri/wannier-berri/blob/f9a947d6a77629295b6fff3887e74fa7056f0818/wannierberri/data_K/data_K.py#L275-L282

stepan-tsirkin (changes requested):
Hi @bowenphys , thank you for the contribution. I cannot check every line now, but the fact that fermi sea and fermi surface implementations give the same is good.  A few things to check:

1) Are the transfromations with TR and Inversion correct? Did you try to integrate with and without use of pointgroup symmetries, do the results match? What are the symmetries o your model?

2) Is this implementation a part of some paper (in preparation/preprint/published)

3) Do you have some notes with the equations on which you base the implementation.

4) Di you test with materials with Kramers degeneracy?  In my understanding, the integrated quantitywill vanisg if the system has either P or T, but if it has neither of them, but has P*T, the net spin-berry dipole is non-zero. So a simple antiferromagnet (like Cr2O3) or a simple model for an antiferromagnet should be suitable. However, locally (at selected k-points) SpinBerry curvature and its derivative are nonzero already for something as simple as Platinum.

wannierberri/formula/covariant.py
bowenphys: I think you are right here. I originally set this by analogy with the existing charge `DerOmega` structure, but since `SpinOmega` itself is `transform_ident` under both TR and inversion, taking the k-derivative should indeed flip both parities. My current tests were all done with `use_irred_kpt=False` and `symmetrize=False`, so they do not check this part. I will revisit this assignment and test it explicitly with symmetry reduction.

wannierberri/formula/covariant.py
```diff
@@ -780,6 +780,104 @@ def ln(self, ik, inn, out):
...
+            dEinv[ik, diag, diag] = 0.0
+
+        delta_v = data_K.delE_K[:, :, None, :] - data_K.delE_K[:, None, :, :]
+        delta_v_over_de = delta_v * dEinv[:, :, :, None]
+        for ik in range(delta_v_over_de.shape[0]):
+            delta_v_over_de[ik, diag, diag, :] = 0.0
```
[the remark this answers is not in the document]
bowenphys: Agreed, same issue here. Since `dEig_inv` already vanishes on the diagonal / degenerate entries, this additional zeroing is redundant.

bowenphys:
> Hi @bowenphys , thank you for the contribution. I cannot check every line now, but the fact that fermi sea and fermi surface implementations give the same is good. A few things to check:
> 
> 1. Are the transfromations with TR and Inversion correct? Did you try to integrate with and without use of pointgroup symmetries, do the results match? What are the symmetries o your model?
> 2. Is this implementation a part of some paper (in preparation/preprint/published)
> 3. Do you have some notes with the equations on which you base the implementation.
> 4. Di you test with materials with Kramers degeneracy?  In my understanding, the integrated quantitywill vanisg if the system has either P or T, but if it has neither of them, but has P*T, the net spin-berry dipole is non-zero. So a simple antiferromagnet (like Cr2O3) or a simple model for an antiferromagnet should be suitable. However, locally (at selected k-points) SpinBerry curvature and its derivative are nonzero already for something as simple as Platinum.

Hi @stepan-tsirkin , thank you very much for the detailed comments.

This implementation is part of my recent work, and I expect to upload the related manuscript to arXiv soon. One motivation for preparing this work was exactly that I needed to compute spin Berry-curvature dipoles (SBCD). I will also include a detailed analytical derivation in the appendix of that work, so I do have notes behind the equations used here.

I have also tested the antiferromagnetic case on my side, and in the symmetric case the integrated result vanishes as expected, so that part seems consistent.

I started working seriously with WannierBerri around last December, and this is my first substantial contribution to the codebase. Since this is my first serious PR to WannierBerri, I now see that some parts were still too rough and not yet in the right shape for the codebase. Your comments made that very clear to me, especially regarding the symmetry/parity handling and some implementation details.

So rather than trying to defend the current version too much, I will go through your comments carefully, fix the issues one by one, and then submit an updated PR. Thanks again for taking the time to point these things out.

stepan-tsirkin:
> 
> So rather than trying to defend the current version too much, I will go through your comments carefully, fix the issues one by one, and then submit an updated PR. Thanks again for taking the time to point these things out.

You may just continue commiting to the current branch, and updating it from the master (I am doing othe PRs in the meantime) . When the PR is accepted I squash merge, so it anyway will appear as a single commit. 

But as you wish. Looking forward to.

bowenphys:
I have updated and pushed the current PR branch.

In this round, I mainly addressed the points you raised:
- revised the TR / inversion transformation assignment in `DerSpinOmegaSimple`
- removed the redundant `parameters.pop(...)` lines
- removed the extra manual zeroing of `dEig_inv` / `delta_v_over_de`
- cached repeated calls inside `nn()`
- rewrote the reduction helper in the test using `alpha_A` / `beta_A`
- replaced the custom generalized-derivative helper with the existing `Matrix_GenDer_ln`
- added an extra TR-symmetric Kane-Mele test, where the integrated spin-BCD vanishes numerically as expected

I also updated the current branch with the latest `upstream/master` and pushed everything to the same PR branch.

There is still one separate issue which is not really about the spin-BCD logic itself: a few reference `.npz` files in the repository differ only by letter case in their filenames, and on my macOS filesystem this creates a case-sensitivity problem at the working-tree level. If needed, I can explain that part separately.

I will also continue testing this implementation on additional models/material systems on my side.

For now, please do not merge this PR yet. I would still like to add one final comparison figure showing the results with point-group symmetries enabled and disabled. After I post that, I think the PR should be in much better shape for merging.

stepan-tsirkin:
> There is still one separate issue which is not really about the spin-BCD logic itself: a few reference `.npz` files in the repository differ only by letter case in their filenames, and on my macOS filesystem this creates a case-sensitivity problem at the working-tree level. If needed, I can explain that part separately.

@bowenphys  we usually test it on Linux systems, so do not notice the problem. But to improve compatibility with win/mac, you may first submit a separate PR, that just renames those files.  I agree that it is better to avoid such collisions in the repository, to avoid confusions. Maybe we can merge that PR before this one.

bowenphys:
Hi @stepan-tsirkin,

I have now completed the symmetry on/off check that was still missing from this PR.

I tested the spin-BCD calculation on my material with and without the symmetry reduction, using the relevant magnetic symmetry setting discussed here. The results are consistent between the two runs, while the symmetry-enabled calculation is noticeably faster. I attached the corresponding comparison figure to document this explicitly.

symmetry_on_off_comparison_dense

So this addresses the remaining point about checking the implementation against the use of point-group symmetries in practice.

Separately, I also opened a small independent PR for the case-insensitive filename collision issue in the reference `.npz` files:
#553

That issue is unrelated to the spin-BCD logic itself, so I handled it separately as suggested.

stepan-tsirkin:
Great. The only thing is that the tests only compare implementations between themselves. Would be good to add a test that compares to the refernece data (add some npz files). 

This can be dene both on the calculator level (like test_run.py) and formula (test_formula.py) 

When comparing fermi-sea vs fermi-surface you are using quite a dense grid (which is really needed, otherwise they will differ a lot), but when comparing to a reference data, a very coarse grid may be used.


I already merged #555 . As I understand, #553 and #554 were a mistake.

wannierberri/formula/covariant.py
```diff
@@ -789,6 +789,115 @@ def ln(self, ik, inn, out):
...
+
+        dEinv = np.array(data_K.dEig_inv, copy=False)
+        delta_v = data_K.delE_K[:, :, None, :] - data_K.delE_K[:, None, :, :]
+        delta_v_over_de = delta_v * dEinv[:, :, :, None]
+
+        S_beta = np.empty_like(dS)
+        V_beta = np.empty_like(dV)
+        for ik in range(S.shape[0]):
+            S_beta[ik] = s_beta_formula.nn(ik, all_bands, all_bands)
+            V_beta[ik] = v_beta_formula.nn(ik, all_bands, all_bands)
```
stepan-tsirkin: This is not how  Matrix_GenDer_ln(ik, inn, out) is supposed to work. `out` should be a complement of `inn`, so passing `all_bands` for both does not make much sense

bowenphys:
Hi @stepan-tsirkin, thank you for pointing this out. I have updated the implementation so that `Matrix_GenDer_ln` is evaluated with `out` as the actual complement of `inn`; the previous use of `all_bands` for both subspaces has been removed.

While adding the requested reference-data tests, the Linux CI exposed an additional gauge-dependence in degenerate subspaces. The derivative of `J / Delta E` had used individual diagonal band velocities, which depend on the chosen basis inside a degenerate subspace. I replaced this with the covariant subspace form

`dX = (dJ - V_in X + X V_out) / Delta E`, with `X = J / Delta E`,

using the full `V.nn()` and `V.ll()` matrices. I also added a deterministic unitary-rotation test for degenerate subspaces; the spin-Berry-curvature derivative trace is now invariant to numerical precision.

The reference-data coverage now includes:
- a formula-level `.npz` regression test for `DerSpinOmegaSimple`;
- calculator-level `.npz` regression tests for both `SpinBerryDipole_FermiSurf` and `SpinBerryDipole_FermiSea` on a coarse grid, following the `test_run.py` framework;
- the separate dense-grid Fermi-surface/Fermi-sea consistency test;
- the TR-symmetric Kane-Mele test with and without symmetry reduction.

The targeted tests pass on both macOS and Linux (Python 3.13), and all PR checks are now green for Python 3.12, Python 3.13, build/coverage, and flake. I have left the review thread unresolved for you to verify.

PR #534, opened 2026-04-07
Add supercell folding for System_R

yaroslav-zhumagulov:
Adds supercell folding for `System_R`.

## Summary
- New module `wannierberri/system/system_supercell.py` with:
  - `fold_system(system_prim, M)` — fold a primitive `System_R` into a supercell `System_R` for an arbitrary integer 3×3 supercell matrix `M`. All R-space matrices present in the primitive system (`Ham`, `AA`, `BB`, `CC`, `SS`, …) are folded into the block structure.
  - `add_scattering(system_sc, V_kk, grid_shape, M)` — add a k-space scattering potential to the supercell Hamiltonian via double FFT + minimum-image folding.
  - `spin_double_system(system)` — expand a spinless `System_R` into a spin-1/2 system (`I_2 ⊗ X`) with on-site Pauli `SS`.
- New `System_R.build_supercell(M, **kwargs)` instance method as a thin wrapper around `fold_system`.
- Exposed `fold_system`, `add_scattering`, `spin_double_system` from `wannierberri.system`.


wannierberri/system/system_supercell.py
```diff
@@ -0,0 +1,510 @@
...
+        nwann,
+        norb_sc,
+        len(iRvec_prim),
+        len(R_sc),
+        list(folded.keys()),
+    )
+    return system_sc
+
+
+def spin_double_system(system, periodic=None):
```
stepan-tsirkin, line 327: This is already available:

https://github.com/wannier-berri/wannier-berri/blob/49661bb2ad0825c5cc5e9ad303cc79c560f9cc60/wannierberri/system/system_R.py#L417-L440

It doubles the spin in "interlace mode" - which is consistent with the convention of pw2wannier90 and modern VASP.

PR #547, opened 2026-05-15
update docs and spglib function

pinegym:
Changes:
- Update spglib calls in `System.set_pointgroup_from_structure`
  - Obtain `cell = self.get_spglib_cell()` and, if `self.magnetic_moments is None`, call `spglib.get_symmetry(cell)`, otherwise call `spglib.get_magnetic_symmetry(cell)`, so that magnetic symmetry (including time-reversal) for magnetic systems is handled automatically with spglib ≥ 2.

- Update docstring and warning messages in `set_pointgroup_from_structure`
  - Remove the old explanation that “for magnetic systems, symmetries involving time reversal are not detected by spglib,” which reflected the previous behavior.

- Clean up comments in `tests/test_point_symmetry.py::test_symmetry_spglib_Fe`
  - Delete the comment that “magnetic symmetries involving time-reversal are not implemented in spglib, so we exclude such symmetries from the generators”, while leaving the actual test logic (e.g. `symmetries_Fe_except_TR`) unchanged.

### DIFF
--- a/tests/test_formula.py
+++ b/tests/test_formula.py
@@ -123,7 +123,8 @@ def check_formula_output():
                 continue
             maxval = np.max(abs(val))
             if maxval < atol_zero:
-                assert np.allclose(val_out, 0, atol=atol_zero), f"{filename} key {k} is expected to be zero, but max value is {maxval} > {atol_zero}"
+                maxval_out = np.max(abs(val_out))
+                assert maxval_out < atol_zero, f"{filename} key {k} is expected to be zero, but max value is {maxval_out} > {atol_zero}"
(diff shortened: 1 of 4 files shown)


pinegym:
There still seem to be some failing tests, but it doesn’t look like they are caused by my changes.

stepan-tsirkin:
> There still seem to be some failing tests, but it doesn’t look like they are caused by my changes.

Yes, they are not related, and quite weird, because they pass in CI, but not in codecov.

In any case, thank you for the contribution, I will take care of that and merge when I have time.

PR #554, opened 2026-06-08
Rename case-colliding formula reference files

bowenphys:
This PR addresses a repository compatibility issue unrelated to the spin-BCD implementation itself.

On case-insensitive filesystems such as the default setups on macOS and Windows, the following reference files collide at the working-tree level because they differ only by letter case:

- `tests/reference/formula/DerMorb.npz`
- `tests/reference/formula/Dermorb.npz`
- `tests/reference/formula/Der2Morb.npz`
- `tests/reference/formula/Der2morb.npz`

To avoid these collisions, this PR:

- renames those four reference files to case-safe names
- adds a small mapping in `tests/test_formula.py` so the corresponding formulas still load the correct reference data

This does not change the formula logic or the stored reference values themselves; it only makes the repository layout safer on case-insensitive filesystems.


bowenphys:
Closing this attempt and recreating the fix from a clean local git checkout to avoid any ambiguity from the earlier API-based edits on case-colliding files.

PR #573, opened 2026-08-05
Added support for space-groups containing TR in EBRsearcher

julienfah:
# Description

For non-magnetic systems, the `EBRsearcher` was not compatible with time-reversal symmetry (type II groups), which was problematic for cases where the DFT is done with TR to reduce the k-grid, leading to a type II group into the `EBRsearcher`. In order to fix this, the `EBRsearcher` now computes the unitary sub-group, and then uses it for the computations. Nothing is changed if there is no time-reversal symmetry, and when there is only unitary constraints are applied. These remain necessary conditions, so the search returns a superset of the valid combinations, no valid projection set is lost. Type III and IV magnetic groups are still rejected.

# Implementation

It was decided not to mutate the symmetrizer's space group (although it is deep-copied), as it would then fail silently if one would want to reuse it later for any other purpose. Instead the unitary indices are carried explicitly on the searcher and passed to `get_all_possible_irreps_conj`, whose new arguments default to the full group

# Changes
- Implementation of type II groups,
- verbose print moved to debug message,
- import of `spgrep.representation` changed to `spgrep.rep.representation` to avoid a deprecation warning,
- Modify the test file to cover the new implementation using the existing diamond data,
- Separate bug fix: in `wannierberri.symmetry.projections.ProjectionsSet`, the `num_wann_per_site` variable was not cleared from cache after merging positions with `join_same_wyckoff`. It turns out that this variable plus 2 others were commented in `clear_cached_properties`, which suggests there was a hidden reason for them not being cleared. Uncommenting fixed this specific bug.


wannierberri/symmetry/projections_searcher.py
```diff
@@ -375,11 +408,55 @@ def highlight(line):
...
+def unitary_spacegroup(sg, isym_unitary):
+    """A copy of `sg` holding only its unitary operations.
+
+    Returns `sg` itself when there is nothing to strip, so a colourless group
+    takes no copy at all.
+    """
+    if len(isym_unitary) == len(sg.symmetries):
+        return sg
+    sg_u = copy.deepcopy(sg)
+    sg_u.symmetries = [sg.symmetries[i] for i in isym_unitary]
```
stepan-tsirkin, line 461: I think, here one needs to clear chached property 'translations_cart', just in case
https://github.com/irreducible-representations/irrep/blob/3df0ca9ea00af4e69c751669e26b8572a489cbc0/irrep/spacegroup.py#L692-L694

wannierberri/symmetry/projections_searcher.py
```diff
@@ -375,11 +408,55 @@ def highlight(line):
...
+    Type II is the only magnetic type containing pure 1' (identity rotation,
+    zero translation, time-reversal): type III combines TR only with
+    non-identity operations, and type IV's antiunitary identity carries a
+    translation. So this separates II from III/IV.
+    """
+    assert hasattr(sg.symmetries[0], "time_reversal")
+    tr = [bool(s.time_reversal) for s in sg.symmetries]
+    unitary = [i for i, t in enumerate(tr) if not t]
+    is_grey = any(
+        t and np.allclose(s.rotation, np.eye(3)) and np.allclose(s.translation, 0)
```
stepan-tsirkin: What if translation is integer-valued?  Should the group be considered as grey in this context?

stepan-tsirkin (changes requested):
looks good, a few small suggestions (see comments)

PR #574, opened 2026-08-06
Read nnkp file with 0 projections

wenchenyue:
Use checkpoint k-point data by default when a chk file is requested, while preserving explicit readnnkp overrides and nnkp behavior for win-only inputs. Add tests for all selection paths.


stepan-tsirkin:
@wenchenyue , thank you for the fix.

In principle, this fix should also be submitted to wannier90io, so that is can read files with empty projections

Regarding the default value, I would still prefer to have only explicit choice to read nnkp or not (no "None" option, because it becomes little intuitive) 

And the default should be to **read the file** (`readnnkp=True`), because it affects not only wanierisation, but also berry connections end other properties, when the mmn, uHu, uIu, sHu and other properties are included. 

I made the corresponding changes and updated the tests

PR #578, opened 2026-09-10
Wrtie epw

stepan-tsirkin:


DjordjeDangic:
Hello Stepan. Does this thing work with EPW? I would like to use wannierisation from WannierBerri which can symmetrize the Wannier functions with frozen window and use it with EPW.

stepan-tsirkin:
Hi @DjordjeDangic ,  as I am not an EPW user, I did not try it yet. So, it would be nice if you take the initiative to test it and report the problems, so that we can converge to a working solution.

DjordjeDangic:
Hi @stepan-tsirkin . I tried it and manage to get it working in my test system (Pb). There was however some tweaks that I had to do (or rather that I asked AI to do it for me). I do not know what would be the best way of reporting these tweaks, especially since they are specifically suited for the EPW version I am using (v6.0 I think).

